PR #4, opened 2023-03-08
Sam tov genetic training

SamTov:
Add Dash to training and deployment options.

## Summary of Additions

* Added training routines directory
* Added Genetic algorithm training routine
* Added test for genetic training
* Added Ensemble training
* Added ensemble training test

## Example

Now you can deploy SwarmRL jobs in parallel locally or on any desk_jobqueue cluster. Noting that the cluster stuff is not tested as we do not have any CI clusters.

Looks like this:

```python
import swarmrl as srl

gym = srl.Gym(...)

routine = training_routine(
            gym,
            get_simulation_runner,  # Function that returns a simulation runner
            number_of_ensembles = 100000
            n_episodes=1000,
            episode_length=20,
            n_parallel_jobs=50,
    )
routine.train_ensembles()
```

In this case, you would run 100000 models 50 at a time. Running the code automatically opens the Dask dashboard so you can watch in real time the progress.

For the genetic training the API is similar with additional arguments for the genetic algorithm component. 
Check the added tests for additional API details until we manage to build some documentation.

One can also use a custom cluster, for example, with SLURM:

```python
import swarmrl as srl
from dask_jobqueue import SLURMCluster

gym = srl.Gym(...)

cluster = SLURMCluster(cores=12, resources={"espresso": 1, "gpu": 4})

routine = training_routine(
            gym,
            get_simulation_runner,  # Function that returns a simulation runner
            number_of_ensembles = 100000
            n_episodes=1000,
            episode_length=20,
            n_parallel_jobs=50,
            cluster=cluster
    )
routine.train_ensembles()
```
The only caveat here is that one must add `espresso` as a resource. For as long as espresso cannot run two system classes in the same kernel, this will be a requirement.


Dalouvid (approved):
Awesome! Love it.

PR #8, opened 2023-03-16
Update readme

### LINKED ISSUE #6: Update readme
Now that we are public and professional, we should have a better welcome page here on github, highlighting what the software can do

SamTov:
## Update readme

Just adding some more info and even fixing the install link in the readme. 

Fix #6 

* Added information to the readme.


christophlohrmann (approved):

README.md
```diff
@@ -0,0 +1,31 @@
...
+directory.
+
+```sh
+   git clone https://github.com/SwarmRL/SwarmRL.git
+   cd SwarmRL
+   pip install .
+```
+
+This will install the package with a reference to the current directory.
+This means that if you change the code, you will not need to re-install it.
```
christophlohrmann, line 26: the pip command misses ``-e``. Also not sure if we need that level of installation instructions in our readme

christophlohrmann (changes requested):

README.md
SamTov: I can just have the pip install line and say this will install it but without the discussion local installation?

christophlohrmann (approved):

PR #9, opened 2023-03-17
Implement PPO

Dalouvid:

* Fixed **PPO** (sum instead of means, added epsilons) = **test**
* Implemented **GAE +test** 
* Wrote a save_memory function in **utils** that can save an arbitrary memory_dict. Checkout the doc strings
* Adjusted the **gym** to initialize Tasks 
* implemented the initialize method in the **task parent class** 
* changed the **losses** (both) to not return a gradient but to just update the model 
* **gradient sensing vision cone** was a while ago.. doesnt matter i guess. 
* compute_actions now returns index + log_probs 
* __call__ of flax_networks return logits


CI/unit_tests/losses/_test_proximal_policy_loss.py
```diff
@@ -1,4 +1,5 @@
-import numpy as np
```
SamTov, line 1: Please rename this to no have the leading underscore. This means it is being ignored by pytest.

CI/unit_tests/losses/_test_proximal_policy_loss.py
```diff
@@ -46,85 +47,69 @@ def test_compute_actor_loss(self):
...
                     ratios * advantages,
                     np.clip(ratios, 1 - epsilon, 1 + epsilon) * advantages,
                 )
-                loss = np.mean(clipped_loss, 0)
-                loss = np.mean(loss) + entropy_coefficient * entropy
+                loss = np.sum(clipped_loss, 0)
+                loss = np.sum(loss) + entropy_coefficient * entropy
                 true_results.append(loss)
 
         for i, loss in enumerate(results):
```
SamTov, line 115: The test looks cool but maybe you can run over what it is doing and why it is a valid test? I don't fully understand everything by just looking at the code and I am curious as to how it works.

CI/unit_tests/value_functions/test_gae.py
```diff
@@ -0,0 +1,25 @@
...
+        )
+        expected_returns = expected_advantages + values
+
+        advantages = gae(rewards, values)
+        returns = gae.returns(advantages, values)
+
+        onp.testing.assert_allclose(
+            advantages, expected_advantages, rtol=1e-4, atol=1e-4
+        )
+        onp.testing.assert_allclose(returns, expected_returns, rtol=1e-4, atol=1e-4)
```
SamTov, line 25: great test!

espresso
SamTov, line 1: Was this deliberate? It is dangerous to update espresso 😉

swarmrl/gyms/gym.py
```diff
@@ -30,7 +31,7 @@ class Gym:
     def __init__(
         self,
         rl_protocols: List[ActorCritic],
-        loss: Loss,
+        loss: Loss = ProximalPolicyLoss,
```
SamTov, line 34: What are the pros and cons of having PPO as the default. One one hand it is the better algorithm but on the other it is much more expensive.  Perhaps having gradient optimisation as the default is the simpler option. What do you think?

swarmrl/sampling_strategies/sampling_strategy.py
```diff
@@ -18,6 +18,8 @@ def compute_entropy(self, probabilities: np.ndarray) -> float:
         probabilities : np.ndarray (n_colloids, n_actions)
                 Probabilities for each colloid to take specific actions.
         """
+        eps = 1e-8
+        probabilities += eps * np.ones_like(probabilities)
```
SamTov, line 22: you can just do `probabilities += 1e-8` it will map it over all dimensions.

swarmrl/tasks/searching/_gradient_sensing_vision_cone.py
SamTov, line 1: This whole task is not updated to the latest API and should be removed.

swarmrl/tasks/task.py
```diff
@@ -27,6 +27,25 @@ def __init__(self, particle_type: int = 0):
         """
         self.particle_type = particle_type
 
+    def initialize(self, colloids: List[Colloid]):
+        """
+        Initialize the task with starting positions of the colloids.
+
+        The parent method will just pass. This is because some observables
```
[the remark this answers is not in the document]
SamTov: Actually good point, why is this new here?

swarmrl/value_functions/generalized_advantage_estimate.py
```diff
@@ -0,0 +1,84 @@
...
+        Call function for the advantage.
+        Parameters
+        ----------
+        rewards : np.ndarray (n_time_steps, n_particles)
+                A numpy array of rewards to use in the calculation.
+        values : np.ndarray (n_time_steps, n_particles)
+                The prediction of the critic for the episode.
+        Returns
+        -------
+        expected_returns : np.ndarray (n_time_steps, n_particles)
```
SamTov, line 50: Aren't these reduced over time steps? Is the shape correct?

SamTov (changes requested):
The PR is nice, lots of great code addition. One big comment is that the recording and PPO could have been done in seperate PRs as this is a lot and hard to stay on track with what is related to what. 

Please go over the comments from myself and @christophlohrmann and we can keep moving through it from there.

swarmrl/gyms/gym.py
christophlohrmann: also from a software standpoint: do you want the loss by default to be the Class PPO or an instance of PPO? In case of the latter, you need braces () to instantiate an object

swarmrl/tasks/searching/_gradient_sensing_vision_cone.py
```diff
@@ -144,8 +150,14 @@ def __call__(self, observable: np.ndarray):
 
         Parameters
         ----------
-        observable : np.ndarray (dim, )
-                Position of the colloid.
+        observable : np.ndarray (dimension, )
+                Observable of a single colloid.
+        colloid : object
```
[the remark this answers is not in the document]
SamTov: The only reason this task is here is so that when Tobi comes back to look at the code he can make sure nothing is lost from here that is in the new vision cones observable. It should basically never be touched and it will not work with the current RL infrastructure.

swarmrl/value_functions/generalized_advantage_estimate.py
```diff
@@ -0,0 +1,84 @@
...
+                A numpy array of advantages
+        values : np.ndarray (n_time_steps, n_particles)
+                The prediction of the critic for the episode.
+        Returns
+        -------
+        expected_returns : np.ndarray (n_time_steps, n_particles)
+                Expected returns for the rewards.
+        """
+        returns = advantages + values
+        return returns
```
SamTov, line 84: why not just:

```python
return advantages + values
```
This could also not be a class method as it doesn't use `self` anywhere. Tough to argue where it should go though as it also won't be called anywhere else. I guess the `__call__` method could always return both? What would that mean for the other value functions?

SamTov (commented):
Just a couple more questions thrown in but its looking nice.

swarmrl/gyms/gym.py
SamTov: Ahh very good point. This should be an instance not a class.

swarmrl/value_functions/generalized_advantage_estimate.py
Dalouvid: - That's definitely more readable. I'll change that 
- Returning both with the call method works perfectly. 

the problem with other value functions is that, PPO does not work using expected returns. its to noisy therefore it always uses GAE.  
The structure of GAE does not fit our structure of SGD or the expected returns. 
I could change Expected Returns and SGD such that we can use GAE in SGD.
That would make it truly modular, although it might never be used in this combination.

Dalouvid (commented):
Sorry guys, I didn't get that one has to finish ones review to comment on things...
Sorry for letting you wait. Most of your inputs are already in the code. :*

CI/unit_tests/losses/_test_proximal_policy_loss.py
SamTov: No problem, thanks!

swarmrl/tasks/searching/_gradient_sensing_vision_cone.py
SamTov: I would leave it where it is, I just want it there for Tobi when he back.

swarmrl/gyms/gym.py
Dalouvid: Yeah so let's put no default in the gym. 
Only put defaults in the losses (GAE and Gumble) 
And make it actual instances.

swarmrl/value_functions/generalized_advantage_estimate.py
Dalouvid: I'd leave it like this and fix SGD/Expected_returns in another branch. 
Especially because it is never used in a mixture that does not work. SGD/GAE or PPO/ExpRet are never used. 
And by putting a GAE and ExpRet as default in PPO and SGD we takle that for now. 
What do you think?

swarmrl/gyms/gym.py
SamTov: There are two points here though, if you want this to be a default, it must be an instance. Otherwise the code won't run.

We can totally put the loss as default in the Gym I am completely for that as it simplifies everything for the user. I would however go with the faster and more basic gradient optimisation.

swarmrl/value_functions/generalized_advantage_estimate.py
SamTov: Sounds good.

swarmrl/gyms/gym.py
SamTov: Resolved in https://github.com/SwarmRL/SwarmRL/pull/9/commits/fc718a42587e05f369b612b618cfc985f4ff3b6e

SamTov (approved):
Great work man :smile:

christophlohrmann (approved):

PR #12, opened 2023-03-28
absolut values in concentration field observable

Koppeprojects:
The colloid observable currently only outputs relative values from the last observation to the current. With this pull request a similar class is added that returns absolute values. The function that returns absolute values doesn't need to be initialized, which results in some unnecessary code in the current implementation. Maybe making the observations relative should be outsourced from this code and made more general for multiple observable types?  
(After this point is cleared, I will implement another but similar (absolute and relative) observable)


swarmrl/observables/concentration_field.py
```diff
@@ -135,3 +135,37 @@ def compute_observable(self, colloids: List[Colloid]):
...
+
+        Parameters
+        ----------
+        index : int
+                Index of the colloid to compute the observable for.
+        colloids : List[Colloid]
+                List of colloids in the system.
+        """
+        reference_colloid = colloids[index]
+        position = onp.copy(reference_colloid.pos) / self.box_length
```
christophlohrmann, line 166: the box length scaling here is weird. It should be the users responsibility to scale the decay-function if they want this behaviour.

christophlohrmann (changes requested):

swarmrl/observables/concentration_field.py
SamTov: This is my fault code-wise but I do agree, we can update these things in all observables / tasks.

Koppeprojects: Is somebody already working on the update for all the observables?

PR #13, opened 2023-03-28
New add colloid method

Koppeprojects:
To the current state it is nor easily possible to place colloids one by one on precise places and orientations. This pull request adds a method for this and rewrites the original way of adding colloids. Be sure to only add colloids one by one if no existing colloid type is given a new radius. Is this condition reasonable. It is there to be compatible with existing colloid radius handling.


swarmrl/engine/espresso.py
```diff
@@ -59,6 +59,13 @@ def _vector_from_angles(theta, phi):
     )
 
 
+def _angles_from_vector(director):
```
christophlohrmann, line 62: move this to utils

christophlohrmann: the random angles, to

christophlohrmann (changes requested):
looks good. small comment and we can merge

SamTov (dismissed):
I think it looks good and have no additional comments to Christoph.

christophlohrmann:
@Koppeprojects can you fix the formatting so we can merge? This is a nice PR, let's not let it go stale

christophlohrmann (approved):

christophlohrmann (approved):

PR #16, opened 2023-04-01
[WIP] Koppe flexible write interval

Koppeprojects:
Due to the contact with the experimentalists the need emerged for a time_slice > write_interval in the espresso engine. The code makes the timing flexible a new logic is implemented. 

Also the add_rod() method returns more values to make it easier handling multiple rods.


Koppeprojects:
Please review the code flow and check if something is not reasonable (for example very slow for edge cases). 
Maybe you can make a suggestion how elegantly get rid of the terms -1e-6 in the if statements. 
I will add a test for this code once it  doesn't change to much.

swarmrl/engine/espresso.py
```diff
@@ -612,15 +651,26 @@ def integrate(self, n_slices, force_model: swarmrl.models.InteractionModel = Non
...
+            # print("n_slices",n_slices)
             self.integration_initialised = True
 
-        for _ in range(n_slices):
+        old_slice_idx = self.slice_idx
+        while self.slice_idx < old_slice_idx + n_slices:
             if (
-                self.system.time
-                >= self.params.write_interval.m_as("sim_time") * self.write_idx
+                self.step_idx * self.params.time_step.m_as("sim_time")
```
christophlohrmann, line 670: get the floats out and express everything in terms of integers (steps_per_slice, steps_per_write, current_step)

swarmrl/engine/espresso.py
```diff
@@ -630,36 +680,40 @@ def integrate(self, n_slices, force_model: swarmrl.models.InteractionModel = Non
...
+            if (
+                self.step_idx * self.params.time_step.m_as("sim_time")
+                >= self.params.time_slice.m_as("sim_time") * self.slice_idx - 1e-6
+            ):
+                self.slice_idx += 1
+                self.manage_forces(force_model)
+
+            # print("write idx, slice idx, step idx",self.write_idx,
+            # self.slice_idx, self.step_idx)
+            time_to_next_write = (
```
christophlohrmann, line 692: you can avoid a lot of this logic by requiring that write_interval and time_slice are multiples of each other (does not matter which one is the smaller one). You always integrate the smaller number and do the more frequent thing of [forces, writing] until you do the less frequent thing

swarmrl/engine/espresso.py
```diff
@@ -612,15 +651,26 @@ def integrate(self, n_slices, force_model: swarmrl.models.InteractionModel = Non
...
+            # print("write_interval", self.params.write_interval.m_as("sim_time") )
+            # print("time_slice", self.params.time_slice.m_as("sim_time") )
+            # print("time_step", self.params.time_step.m_as("sim_time") )
+            # print("write idx, slice idx, step idx",self.write_idx, self.slice_idx,
+            # self.step_idx)
+            # print("n_slices",n_slices)
             self.integration_initialised = True
 
-        for _ in range(n_slices):
+        old_slice_idx = self.slice_idx
```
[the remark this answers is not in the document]
Koppeprojects: Because I don't know that to each loop iteration also belongs a slice iteration. This is the case in the current code and also if the write_ interval is separated to the slice_ time by a multiplicative factor.

swarmrl/engine/espresso.py
Koppeprojects: nice hint

christophlohrmann (commented):
new implementation looks very reasonable. just need a writing test that checks that both cases (slice larger or smaller than snapshot) work well

CI/espresso_tests/unit_tests/test_integration.py
```diff
@@ -0,0 +1,227 @@
...
+class Process(multiprocessing.Process):
+    """
+    Process class for use in multi simulation testing.
+    """
+
+    def __init__(self, *args, **kwargs):
+        """
+        Multiprocessing class constructor.
+        """
+        multiprocessing.Process.__init__(self, *args, **kwargs)
```
christophlohrmann, line 24: This test is wild. but if it works it works

christophlohrmann (commented):
update the error msg and we are good to go

CI/espresso_tests/unit_tests/test_integration.py
Koppeprojects: It is based on a hint from Sam. I could integrate over all the process calls from one function and have several simulation classes as well but the Simulations are similar so that maybe it is ok to stick to this. (I try to have rather little  code duplication but still readable code.) This test structure is also implemented in the visualization branch.

swarmrl/engine/espresso.py
```diff
@@ -178,6 +180,15 @@ def _init_system(self):
         self.system.time_step = time_step
         self.system.cell_system.skin = 0.4
 
+        # set writer params
+        steps_per_write_interval = int(round(write_interval / time_step))
+        self.params.steps_per_write_interval = steps_per_write_interval
+        if abs(steps_per_write_interval - write_interval / time_step) > 1e-10:
+            raise ValueError(
+                "inconsistent parameters: time_slice must be integer multiple of"
```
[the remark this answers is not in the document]
Koppeprojects: thank you

christophlohrmann (approved):

PR #20, opened 2023-04-18
[WIP] Update bug fixes

SamTov:
Fix some bugs in the RL code.

* Removed self-sensing from particle sensing
* Added test for initialisation to be sure that multiple initialisations do not change the state as it should not.
* Checked that there is no cyclic changing of observables, i.e, fixed really dumb bug I wrote into the code the first time around.

### DIFF
--- a/swarmrl/observables/particle_sensing.py
+++ b/swarmrl/observables/particle_sensing.py
@@ -80,12 +80,12 @@ class ParticleSensing(Observable):
             indices.append(colloids[index].id)
             positions.append(colloids[index].pos)
 
-        test_points = np.array(
+        sensed_colloids = np.array(
             [colloid.pos for colloid in colloids if colloid.type == self.sensing_type]
         )
 
         out_indices, _, field_values = self.observable_fn(
-            np.array(indices), np.array(positions), test_points, historic_values
+            np.array(indices), np.array(positions), sensed_colloids, historic_values
         )
 
         for index, value in zip(out_indices, onp.array(field_values)):
--- a/swarmrl/observables/particle_sensing.py
+++ b/swarmrl/observables/particle_sensing.py
@@ -53,7 +53,9 @@ class ParticleSensing(Observable):
         self.historical_field = {}
 
         self.observable_fn = jax.vmap(
-            self.compute_single_observable, in_axes=(0, 0, None, 0)
+            self.compute_single_observable,
+            in_axes=(0, 0, None, 0),
+            # out_axes=()
         )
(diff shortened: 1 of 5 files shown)


christophlohrmann (commented):
seems reasonable. Can you update the tests to reflect the changes?

SamTov:
> seems reasonable. Can you update the tests to reflect the changes?

Yep! I will fix the tests and then it should be okay for merge.

SamTov:
@christophlohrmann I have added tests to the observable and task that ensure that when the observable is called many times without changes that it also does not change its value.

christophlohrmann (dismissed):

christophlohrmann:
maybe @Dalouvid should have a look at this too before we merge

SamTov:
> maybe @Dalouvid should have a look at this too before we merge

Sounds good! There is no rush.

swarmrl/tasks/searching/species_search.py
Dalouvid, line 1: "Observable" in the task is kinda misleading and could cause confusion.

swarmrl/observables/particle_sensing.py
```diff
@@ -82,7 +84,7 @@ def initialize(self, colloids: List[Colloid]):
             [colloid.pos for colloid in colloids if colloid.type == self.sensing_type]
         )
 
-        out_indices, field_values = self.observable_fn(
+        out_indices, _, field_values = self.observable_fn(
             np.array(indices), np.array(positions), test_points, historic_values
```
Dalouvid, line 88: The variable name "test_points" was not clear to me. One could think about renaming that to a more intuitive name. Something like "sensing_points", "sensor_points" or "sensed_points"

Dalouvid (changes requested):
Nice job. Except for the mentioned point, super clean.

swarmrl/observables/particle_sensing.py
SamTov: Resolved in https://github.com/SwarmRL/SwarmRL/pull/20/commits/63ab2e0d3b0653d9db8cfd473644b14e7a887a3c

swarmrl/tasks/searching/species_search.py
SamTov: Resolved in https://github.com/SwarmRL/SwarmRL/pull/20/commits/63ab2e0d3b0653d9db8cfd473644b14e7a887a3c

Dalouvid (approved):
Looks good!

PR #22, opened 2023-05-21
Velocity in colloids

### LINKED ISSUE #21: Add velocity to interaction model colloids
**What feature would you like to see added?**
It would be nice to have velocity as a property in the interaction model colloids.

SamTov:
Fix #21 

Added velocity to the Colloid in interaction model and returned them in the espresso simulation.

### DIFF
--- a/CI/unit_tests/interaction_models/test_ml_model.py
+++ b/CI/unit_tests/interaction_models/test_ml_model.py
@@ -126,9 +126,15 @@ class TestMLModel:
         """
         Test species and paricle actions are returned correctly.
         """
-        colloid_1 = Colloid(np.array([3, 7, 1]), np.array([0, 0, 1]), 0, 1)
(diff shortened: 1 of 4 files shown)


christophlohrmann (dismissed):

Koppeprojects:
I think with adding the velocity to the Colloid object the arguments that were meant to be colloid.type were interpreted as colloid.velocities. Because in the tests that I fixed the argument order mattered.

SamTov:
> I think with adding the velocity to the Colloid object the arguments that were meant to be colloid.type were interpreted as colloid.velocities. Because in the tests that I fixed the argument order mattered.

That's a great catch!

Koppeprojects:
But If the code silently accepts single numbers instead of numpy.arrays with 3 entries then maybe bug fixing in the future will be more difficult. Should we generate an error here, or does it restrict the usage to much?

PR #27, opened 2023-07-05
PR 1: Change data handling

SamTov:
* Removed all hidden files involved in the rl training in favour of a data class handler for trajectory data, speedup (1 second per episode) due to drastically reduced read/write.
* While changing this I found one place for a JIT compile so I added that

### DIFF
--- a/swarmrl/gyms/gym.py
+++ b/swarmrl/gyms/gym.py
@@ -207,10 +207,8 @@ class Gym:
         progress = Progress(
             "Episode: {task.fields[Episode]}",
             BarColumn(),
-            (
-                "Episode reward: {task.fields[current_reward]} Running Reward:"
-                " {task.fields[running_reward]}"
-            ),
+            "Episode reward: {task.fields[current_reward]} Running Reward:"
+            " {task.fields[running_reward]}",
             TimeRemainingColumn(),
         )
 
--- a/swarmrl/gyms/gym.py
+++ b/swarmrl/gyms/gym.py
@@ -1,7 +1,6 @@
 """
 Module to implement a simple multi-layer perceptron for the colloids.
 """
-import os
 from typing import List, Tuple
 
 import numpy as np
@@ -79,7 +78,7 @@ class Gym:
             actions=actions,
         )
 
-    def update_rl(self) -> Tuple[MLModel, np.ndarray]:
+    def update_rl(self, trajectory_data: dict) -> Tuple[MLModel, np.ndarray]:
         """
         Update the RL algorithm.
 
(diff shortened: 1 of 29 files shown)


swarmrl/losses/proximal_policy_loss.py
```diff
@@ -205,11 +205,10 @@ def compute_loss(self, actor: FlaxModel, critic: FlaxModel, episode_data):
         model_tuple : tuple  FlaxModel, FlaxModel
             The updated actor and critic network.
         """
-        feature_data = episode_data.item().get("features")
-        old_log_probs_data = episode_data.item().get("log_probs")
-        action_data = episode_data.item().get("actions")
-        # will return the reward per particle.
-        reward_data = episode_data.item().get("rewards")
+        old_log_probs_data = episode_data.log_probs
+        feature_data = np.array(episode_data.features)
```
Dalouvid, line 209: does this has to be converted into an array? 
Especially for the graphs, it would be cool if this can stay flexible.

swarmrl/models/ml_model.py
```diff
@@ -96,15 +108,13 @@ def calc_action(
...
-                record_trajectory(
-                    particle_type=item,
-                    features=np.array(observables[item]),
-                    actions=np.array(action_indices[item]),
-                    log_probs=np.array(log_probs[item]),
-                    rewards=np.array(rewards[item]),
-                )
-        return actions
+        # if self.record_traj:
+        for item in self.particle_types:
```
Dalouvid, line 114: I've noticed that using `type_` as iteration variable instead of `item` makes the code easier to understand for me. 
How do you feel about that?

swarmrl/gyms/gym.py
```diff
@@ -237,11 +239,4 @@ def perform_rl_training(
 
         system_runner.finalize()
 
-        # Remove the file at the end of the training.
-        for item in self.rl_protocols:
-            try:
-                os.remove(f".traj_data_{item}.npy")
-            except FileNotFoundError:
-                pass
-
```
Dalouvid, line 246: Where is the `trajectory_data` cleared for the next episode?

Dalouvid (commented):
I like the idea of the data class!

swarmrl/losses/proximal_policy_loss.py
SamTov: No need for an array at all actually, it should come out as one. I will make this change.

SamTov: Before I change out the numpy arrays, which are currently required for the code to work, I think we should wait until we have a case for graphs during which we can adapt the loss to handle this generic pytree input.

christophlohrmann (approved):

PR #29, opened 2023-07-25
(Discussion): Episodic training

Dalouvid:
A method to reset the system after each episode. 
It can be activated in the gym by setting the "reset" flag to true. 

* added a reset function in the espresso engine 
* added a test for that

``` run the tests i guess. 

```

### DIFF
--- a/CI/espresso_tests/integration_tests/test_episodic_training.py
+++ b/CI/espresso_tests/integration_tests/test_episodic_training.py
@@ -204,14 +204,7 @@ class TestRLEpisodicScript(ut.TestCase):
             episodic_training=True,
         )
 
-    def test_reset(self):
-        # test if the system is reset and the number of colloids is the same
-        num_of_cols_old = len(self.system_runner.colloids)
-        self.system_runner.reset_system()
-        num_of_cols_new = len(self.system_runner.colloids)
-        self.assertEqual(num_of_cols_old, num_of_cols_new)
-
-    def test_add_colloids(self):
+    def test_resets(self):
         # add a different type of colloids and check if the number of colloids is
         # correct
         # also check if the ids and the types are correct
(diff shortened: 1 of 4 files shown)


christophlohrmann (changes requested):
The reset includes a lot of coupling for all functions that add particles and have to remember to store their arguments.
The reset doesn't reset system time nor any part of the trajectory writing.

I am still in favour of handling system resets in the script that sets up the system, i.e. creating a fresh engine after each reset and only reusing the espresso system that's inside the engine

Dalouvid:
> The reset includes a lot of coupling for all functions that add particles and have to remember to store their arguments. The reset doesn't reset system time nor any part of the trajectory writing.
> 
> I am still in favour of handling system resets in the script that sets up the system, i.e. creating a fresh engine after each reset and only reusing the espresso system that's inside the engine

How about we do it like this until the global variables are gone from espresso?
JN will work on it in the next Espresso Coding day so it is only a matter of time until it is done.
When this is done, we can have a clean way of resetting the system.

SamTov:
> > The reset includes a lot of coupling for all functions that add particles and have to remember to store their arguments. The reset doesn't reset system time nor any part of the trajectory writing.
> > I am still in favour of handling system resets in the script that sets up the system, i.e. creating a fresh engine after each reset and only reusing the espresso system that's inside the engine
> 
> How about we do it like this until the global variables are gone from espresso? JN will work on it in the next Espresso Coding day so it is only a matter of time until it is done. When this is done, we can have a clean way of resetting the system.

I'm not sure I fully understand the comment here. Do you mean we do it the way @christophlohrmann says or the PR? If you are referring to the PR, then I would be against merging it into the main codebase with the intention of later removing it. In this case, I would suggest that those who need it keep maintaining the branch and use it until there is a permanent solution.

PR #45, opened 2023-09-15
changed flax_network call method to operate on whole epoch data.

Dalouvid:
Fix #

* the call method of a flax newtwork now uses a vectorized version of the apply_fn. It can operate on an array of feature data to compute logits/values for a whole episode
* in the loss, the vectroized versions are called

```

```


SamTov (approved):
Looks great! Thanks for the quick changes

PR #49, opened 2023-10-06
[WIP] Add checkpointing

### LINKED ISSUE #48: Checkpointing RL models
**What feature would you like to see added?**
For very long runs or when clusters / computers shut down, it would be great to have some model checkpointing available.

**Describe how this feature would improve the code**
It would make the code more resilient to spontaneous events like shutdowns but would also be convenient in the rare case where someone forgets to save their models.

**Describe how this feature would work/possible implementations**
We already have model saving built into the code. We can add a checkpointing argument to the `Gym`, which calls this to a special directory called `ckpts`.

SamTov:
Fix #48 by adding a checkpointing callback to the code

* Moved model saving to the protocols
* Updates some type hints and comments
* Made the paths in the model saving OS independent
* Added callback module
* Added a uniform checkpoint callback


CI/unit_tests/callbacks/test_loss_callbacks.py
Dalouvid, line 1: do we also have a test for the max_reward_checkpointing?

PR #50, opened 2023-10-18
Add flow, potential, langevin

christophlohrmann:
- Add external potential to simulations via interpolation on grid
- Add flow to simulations via interpolation on grid
- Choose langevin or brownian thermostat (Brownian has random velocities, can't be used with flow field)
- Add test that covers all the tricks necessary to run a simulation of particles constrained in a geometry with flow

### DIFF
--- a/.github/workflows/espresso.yml
+++ b/.github/workflows/espresso.yml
@@ -38,7 +38,7 @@ jobs:
           git clone https://github.com/espressomd/espresso.git
           cd espresso
           mkdir build && cd build
-          cp ../../myconfig.hpp .
+          cp ../maintainer/configs/maxset.hpp myconfig.hpp .
           cmake ../ -D ESPRESSO_BUILD_WITH_FFTW=OFF
           cmake --build . -j 5
       - name: Run test suite
--- a/.github/workflows/espresso.yml
+++ b/.github/workflows/espresso.yml
@@ -38,7 +38,7 @@ jobs:
           git clone https://github.com/espressomd/espresso.git
           cd espresso
           mkdir build && cd build
-          cp ../maintainer/configs/maxset.hpp myconfig.hpp .
+          cp ../maintainer/configs/maxset.hpp myconfig.hpp
           cmake ../ -D ESPRESSO_BUILD_WITH_FFTW=OFF
           cmake --build . -j 5
       - name: Run test suite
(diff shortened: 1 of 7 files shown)


swarmrl/engine/espresso.py
```diff
@@ -48,6 +52,7 @@ class to hold all information needed to setup and run the MD simulation.
     time_step: pint.Quantity
     time_slice: pint.Quantity
     write_interval: pint.Quantity
```
SamTov, line 54: I guess periodicity should also go in here. We can save this for a different PR.

.github/workflows/espresso.yml
```diff
@@ -38,7 +38,7 @@ jobs:
           git clone https://github.com/espressomd/espresso.git
           cd espresso
           mkdir build && cd build
-          cp ../../myconfig.hpp .
+          cp ../maintainer/configs/maxset.hpp myconfig.hpp
```
SamTov, line 41: We should probably document at some stage how people should install espresso within SwarmRL. We could try a script approach. Maybe we can open a discussion about it.

SamTov (dismissed):
The code is well written and tested. I have small comments scattered around but nothing critical so I will approve here.

CI/espresso_tests/integration_tests/test_genetic_deployment.py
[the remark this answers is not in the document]
christophlohrmann: don't thank me too early. The test fails for reasons that are beyond me

.github/workflows/espresso.yml
christophlohrmann: maybe we just extract the install that CI uses, put it in a bash script and place it next to the requirements.
Different PR though

SamTov (approved):
No updates since the last review so I still approve

SamTov (approved):
Still fine

christophlohrmann:
@SamTov Thanks very much for pushing this over the finish line.
The fact that the tests fail from espresso duplication errors just because some numbers are changed hints at some deeper problems, though

SamTov:
> @SamTov Thanks very much for pushing this over the finish line. The fact that the tests fail from espresso duplication errors just because some numbers are changed hints at some deeper problems, though

@christophlohrmann There are some deeper issues there. I do sort of know why the tests fail though. The Dask errors can arise when the jobs they are running are so short that there is seemingly not enough time to properly close and reopen a worker which leads to a new script being run on the old one which dies due to the `cannot have two instances of the system class` issue. The other problem was related more to Jax than Dask. These answers aren't fully acceptable as it is impossible to know beforehand if this will happen, but the issue isn't completely unknown.

PR #54, opened 2023-10-25
Code format rule updates

SamTov:
The formatting tool black has changed its policy about new lines under comments. This means most files are being changed here.

In addition to that, I removed a line from `gym.py` and changed the versions of black and isort in the `lint.yaml` file. In essence, this PR is a formatting one. If we don't want it right now we can ignore the specific `black` error and update later... or not at all. I will leave it here for disussion.


christophlohrmann (approved):
I say get it in to make CI go back to normal

PR #56, opened 2023-11-03
Fix the two training routines in SwarmRL

### LINKED ISSUE #41: Ensemble training is no longer working (Now also genetic training)
The ensemble training is no longer working. In the CI we get a pickle error and in local tests on MacOS I receive a `too many open files` error when only two should open (far smaller than the 256 limit for the OS).

SamTov:
Fix #41 

This PR aims to resolve the issues arising in the Dask-related training routines. On the local side, only the ensemble training had minor problems related to recent code changes. But as they had been removed for some time, this was expected. Otherwise, no issues were running the tests locally. ~~Therefore, I assume this is a problem with the available resources on the GitHub runners. To remedy this, I have made the tests smaller.~~ This was not the problem but making the tests smaller has improved the run times so I will leave it. It looks like the problem came from setting the number of threads in dask to 1. Any number great than 1 will work. 

* Fix formatting issues with ensemble training
* Put genetic training test into a `tempdir`
* Reduce the size of the tests, i.e. number of episodes, episode length and so on.

You can run either the genetic algorithm or ensemble training test locally, but as mentioned, this did not cause issues in the first place. I could not reproduce the problem on my machine.

### DIFF
--- a/CI/espresso_tests/integration_tests/test_rl_script.py
+++ b/CI/espresso_tests/integration_tests/test_rl_script.py
@@ -118,7 +118,7 @@ class TestRLScript(ut.TestCase):
         network = srl.networks.FlaxModel(
             flax_model=acto_critic,
             optimizer=optax.adam(learning_rate=0.001),
-            input_shape=(1,),
+            input_shape=(1, 1),
             sampling_strategy=sampling_strategy,
             exploration_policy=exploration_policy,
         )
--- a/CI/espresso_tests/integration_tests/test_rl_script.py
+++ b/CI/espresso_tests/integration_tests/test_rl_script.py
@@ -119,7 +119,7 @@ class TestRLScript(ut.TestCase):
         network = srl.networks.FlaxModel(
             flax_model=acto_critic,
             optimizer=optax.adam(learning_rate=0.001),
-            input_shape=(1, 1),
+            input_shape=(1,),
             sampling_strategy=sampling_strategy,
             exploration_policy=exploration_policy,
         )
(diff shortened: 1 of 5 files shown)


christophlohrmann (approved):
if the tests go green, I have nothing to complain

SamTov:
I have a very fun bug here... When I run the tests inside my vscode terminal, they pass, and when I use the non-vscode therminal, they fail.

SamTov:
> I have a very fun bug here... When I run the tests inside my vscode terminal, they pass, and when I use the non-vscode t terminal, they fail.

Not that I can tell you why, but on some systems, setting the number of cores for the ML stuff to 1 killed everything. From what I can understand, they all started to try and run on the same cores and just kept failing, opening files, and eventually, overloadig the OS. Setting `threads=2` resolves the issue completely.

CI/espresso_tests/integration_tests/test_rl_script.py
```diff
@@ -119,7 +119,7 @@ def __call__(self, x):
         network = srl.networks.FlaxModel(
             flax_model=acto_critic,
             optimizer=optax.adam(learning_rate=0.001),
-            input_shape=(1,),
+            input_shape=(1, 1),
```
Dalouvid, line 122: Do we actually have to specify the shape like that?

Dalouvid (commented):
All fine but this one question about the shape.

CI/espresso_tests/integration_tests/test_rl_script.py
SamTov: no, that should be gone actually, the one above is correc.t

Dalouvid (approved):

CI/espresso_tests/integration_tests/test_rl_script.py
SamTov: Oh but this was fixed in the next pr... I just had this in here to see if that was the problem.

christophlohrmann (approved):

PR #57, opened 2023-11-06
PGO Debugging

SamTov:
We have noticed after the recent model API change that the simple policy gradient updates are struggling to train. I am going through each stage and cleaning up the data transport to make this easier. A collection of my changes are below:

* Standardized `vmap` calls on model input. Now, users should provide only the observable shape to the network and not the batch dimension. During action calls this has a single vmap over agents and during training it has two vmaps over agents and time steps.
* Standardized the operations in losses. We now sum over time and agents. I fixed the code comments and shortened it to `.sum()` operations.

The problem with PGO was simply the reward magnitude being too small. I have successfully trained chemotaxis in 200 steps.

### DIFF
--- a/CI/unit_tests/networks/test_flax_network.py
+++ b/CI/unit_tests/networks/test_flax_network.py
@@ -58,7 +58,6 @@ class TestFlaxNetwork:
             exploration_policy=self.exploration_policy,
         )
         input_data = np.array([[1.0, 2.0], [4.0, 5.0]])
-        print(input_data.shape)
 
         data_from_call, value_from_call = model.apply_fn(
             {"params": model.model_state.params}, input_data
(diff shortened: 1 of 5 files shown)


Dalouvid (dismissed):

Dalouvid (changes requested):
I think there is something wrong with the test. 
This does not actually test the call function.

input_data would be all observations of the swarm for a single time step. 
this is an input for the compute_action function
(num_agents, obs_dim)

the call_method is there to be called on a batch of these observations of a whole episode. 
Therefore the input to the call function has to have one more dimension. 
(episod_length, num_agents, obs_dim)
And then we have to actually call the __call__ method, not apply_fn. 

PS: 
in the call method, batch_apply_fn of the network is called, in the compute_action, the apply_fn.

SamTov:
> I think there is something wrong with the test. This does not actually test the call function.
> 
> input_data would be all observations of the swarm for a single time step. this is an input for the compute_action function (num_agents, obs_dim)
> 
> the call_method is there to be called on a batch of these observations of a whole episode. Therefore the input to the call function has to have one more dimension. (episod_length, num_agents, obs_dim) And then we have to actually call the **call** method, not apply_fn.
> 
> PS: in the call method, batch_apply_fn of the network is called, in the compute_action, the apply_fn.

I think the view here is the wrong way around, but yes, there is not test of the `call` method here. The way it is now is as follows:

* `apply_fn` is mapped over agents only and is used in the action calculation
* The `call` method has the additional vmap on it so it maps over time as well

By keeping them this way, the dimension of the observable alone is always passed into the network which is what we want to have.  In the tests I changed here, there is no time given which means you cant use the `call` method.

We could add a new test to check the `call` method but in the case above, the tests are correct, they just test a different method.

SamTov:
I added a new test to check the `call` method specifically.

PR #59, opened 2023-11-09
espresso engine: add system reuse

christophlohrmann:
You can now recycle the system created by one engine to power another engine.
Having multiple parallel engines isn't possible, but you can do serial engine creations like
```python
engine_0 = EspressoMD(...)
engine_1 = EspressoMD(..., system=engine_0.system)
engine_0 = None
...
```
In the above example, engine_0 is compromised after the creation of engine_1, because the latter is writing stuff to the system that is still part of engine_0. I therefore explicitly delete engine_0.

This is not the cleanest solution to the "only one system"-Problem, but imo it is much safer than writing a reset for the swarmrl engine itself.


christophlohrmann:
the coding style issue is annoying. I re-initialised my pre-commit environment before commiting, it should have the correct black version

SamTov:
The dask tests are also annoying... I will try some other workarounds

SamTov:
In this framework, could I do the following:

```python

engine_0 = srl.EspressoMD(...)

for episode in n_episodes:
    engine_1 = EspressoMD(..., system=engine_0.system). # Set engine_1 to initial condition (random placing of particles)
    engine_1.run(1000). # Run for 1000 timesteps and create trajectory
```
In these cases, will the system reset each time back to its initial conditions? I don't understand how this can be possible.

christophlohrmann:
```python
def setup_engine(system):
    engine = EspressoMD(..., system=system)
    engine.add_particles(...) #and all the other steup you need to do
    return engine

system = espressomd.System()

for i in range(100):
    engine = setup_engine(system)
    engine.run(...)
```

SamTov (approved):
Simple and elegant, just how a PR should be.

PR #60, opened 2023-11-09
unite pre-commit and CI linting

christophlohrmann:
... to make sure that the two always use the same version of black, isort, ...

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -2,9 +2,10 @@ name: Check coding style
 
 on:
   push:
+  pull_request:
 
 jobs:
-  main:
+  precommit:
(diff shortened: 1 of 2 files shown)


SamTov (approved):
Happy with it

christophlohrmann:
don't merge. This isn't working correctly yet

SamTov:
> should we keep the auto-fix or should I try to disable it?

I actually like it.

SamTov:
The linting rules need to be removed from the branch rules for this to pass I think.

christophlohrmann:
then this is good to go once it's all green

SamTov:
> The linting rules need to be removed from the branch rules for this to pass I think.

Took care of it

PR #62, opened 2023-11-12
Implement Episodic training

### LINKED ISSUE #55: Adjustment to the Gym
Based on discussion last week, we feel the `Gym` would be best restructured to include the style of training that is being performed. The first change is changing the name to `Trainer` as a parent class. Then, we thought of the following trainers:

* `ContinuousTrainer`: Train the RL models continuously
* `EpisodicTrainer`: Using episodic training
* `Genetic(Continuous/Episodic)Trainer`: Use the genetic algorithm to iteratively select the best models
* `Ensemble(Continuous/Episodic)Trainer

This would integrate the training routines directly into  the code and be more inline with software standards.

SamTov:
Implement episodic training using new methods for re-using systems.

closes #55 

Changes made:

- [x] Implement episodic training class 
- [x] Implement semi-episodic training class
- [x] Add `end simulation` flag in espresso runner and RL tasks
- [ ] ~~Consolidate data output from episodic runs~~
- [x] Write tests for everything and pray they pass.

Consolidation of the files is quite complex in its current state as it is on the user's side. We could add it to the main parts of the code, but then we would need to change other modules that create directories, and, at the end of it all, a user is not protected from messing up the internals. We need to discuss how to store trajectories moving forward for all cases. I have added the issue #63 to address this.


SamTov:
After good conversations with @christophlohrmann, the kill switch is not a property of the force model which means no return is required in the interaction models and all API will remain constant.

swarmrl/training_routines/ensemble_submit.py
Dalouvid, line 1: Why have a "gym" in the init and not "trainer"?

swarmrl/training_routines/genetic_algorithm.py
Dalouvid, line 1: same as in ensemble_submit

CI/unit_tests/interaction_models/test_bechinger_models.py
```diff
@@ -34,6 +34,7 @@ def test_force(self):
         colloids = [test_coll, coll_front, coll_back, coll_side]
 
         action = self.force_model.calc_action(colloids)
+        print(action)
```
Dalouvid, line 37: Leftover?

Dalouvid (commented):
No functional questions. Just the few questions.

swarmrl/engine/espresso.py
```diff
@@ -1011,7 +1011,24 @@ def _remove_overlap(self):
...
+
+        Parameters
+        ----------
+        force_model : swarmrl.models.InteractionModel
+            Model with which to compute external forces.
+
+        Returns
+        -------
+        kill_switch : bool
+            A boolean flag that ends the simulation early.
```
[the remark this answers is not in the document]
SamTov: Yes, this comment should be removed, good catch.

CI/unit_tests/interaction_models/test_bechinger_models.py
SamTov: Yep!

Dalouvid (approved):

PR #64, opened 2023-11-26
Introducing Agents (PR 2)

### LINKED ISSUE #39: ML-Model Data handler is never cleared in deployment mode
**Describe the bug**
The ML Model Data Handler is never cleared in deployment mode. 
During training this is taken care of by passing a new data handler after an update. When deploying a model as a force_fn this will never be done. 

**To Fix**
Just activate the usage of the "record_training" flag in the ML model and set it to true by default.

SamTov:
Introduce the agent class to SwarmRL. The agent class can be RL or classical and talks to the ForceFunctions. This allows for uniformity between API when using classical or RL algorithms and allows us to easily mix the two.

The following has been implemented:

- Create agent class
- Move RL and Classical algortihms to agents
- Create unified bridge for  agents and the system runner allow classical and RL to be put together
- Move data collection to the agents, not in the force functions, closes #39 
- Put actions into their own class as there may be new ones soon (Globals)
- Check for remaining Colloid calls to the wrong place and correct them
- Update the tests for the new structure


swarmrl/force_functions/force_fn.py
```diff
@@ -0,0 +1,87 @@
...
+
+import numpy as np
+
+from swarmrl.actions.actions import Action
+from swarmrl.components.colloid import Colloid
+
+
+class ForceFunction:
+    """
+    Class for a NN based espresso interaction model.
```
Dalouvid, line 15: "shit I forgot to fix the coment"

swarmrl/force_functions/force_fn.py
```diff
@@ -0,0 +1,87 @@
...
+        Constructor for the NNModel.
+
+        Parameters
+        ----------
+        agents : dict
+            Agents used in the simulations.
+        """
+        super().__init__()
+        self.agents = agents
+        self.eps = np.finfo(np.float32).eps.item()
```
Dalouvid, line 34: kill that

christophlohrmann (approved):
change the function name pls, otherwise good

PR #73, opened 2023-12-19
Add lb 

christophlohrmann:
Also add default values to the espresso params dict and to some methods.
Maybe @Dalouvid and @Koppeprojects have an opinion on the fefault values that I chose


SamTov (approved):
Looks great

PR #77, opened 2024-02-14
adapt to new espresso virtual sites handling

christophlohrmann:


christophlohrmann:
@SamTov do you know where the jax problem with the key array comes from?
Seems like a dependency version update. Should we fix the Jax version in ``requirements.txt``?

SamTov:
> @SamTov do you know where the jax problem with the key array comes from? Seems like a dependency version update. Should we fix the Jax version in `requirements.txt`?

Yeah we just saw this today, I will do some digging and get this sorted out by tomorrow and fix a version.

SamTov:
The issue is not actually Jax but a package called Chex. It seems like this particular case is a result of the numpy dependency. I hope it is resolved with the changes in the requirements. 

I am not a huge fan of fixing the Jax libraries, in general, it is quite hard to get a Jax version that works for everyone, particularly if they want a certain GPU compatability. I don't know the best solution to this problem but perhaps we can fix it and then provide extensive documentation. Or, we go for a more intelligent solution after some research.

SamTov:
The error in the espresso test here appears to be still present. This is different to the install error and seems like a possible virtual site thing.

SamTov (approved):
As long as tests are repaired, I am very happy.

PR #78, opened 2024-02-14
DaZi graph agents - NOT WORKING

Dalouvid:
Translated the Attentive Graph Obst to the new Agent-structure.

### DIFF
--- a/swarmrl/observables/col_graph.py
+++ b/swarmrl/observables/col_graph.py
@@ -1,0 +1,243 @@
+from typing import Any, Iterable, List, Mapping, NamedTuple, Optional, Union
+
+import jax.numpy as np
+import numpy as onp
+from jax import config, jit, vmap
+
+from swarmrl.components import Colloid
+from swarmrl.observables.observable import Observable
+from swarmrl.utils.utils import calc_signed_angle_between_directors
+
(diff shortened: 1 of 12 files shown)


swarmrl/observables/col_graph.py
```diff
@@ -0,0 +1,243 @@
...
+        nodes_list = []
+        destinations_list = []
+        receivers_list = []
+        senders_list = []
+        n_nodes_list = []
+        n_edges_list = []
+        globals_list = []
+
+        for col in colloids:
+            if col.type == self.particle_type:
```
SamTov, line 135: I added some helper methods to collect these indices without the need for an if statement.

swarmrl/observables/col_graph.py
```diff
@@ -0,0 +1,243 @@
...
+
+        nodes_list = []
+        destinations_list = []
+        receivers_list = []
+        senders_list = []
+        n_nodes_list = []
+        n_edges_list = []
+        globals_list = []
+
+        for col in colloids:
```
SamTov, line 134: This would be greatly accelerated if it was a vmap over colloid objects.

SamTov (commented):
I just had quick look over and added two comments for future work into the branch. In general, I have to say, the code is quite well written. By looking at it shortly, I can see a few places where any JIT would greatly diminish first episode performance, something we can probable fix.

PR #82, opened 2024-02-22
Resobee engine

Mag-r:
Fix #

* Updated ResoBee integration to work with latest state of SwarmRL
* 

```

```


christophlohrmann:
please point the PR at the main branch so we can see what is actually new

SamTov:
For additional information, there are additions in this code that are incompatible with the ESPResSo engine. I will need to resolve these in a different branch before this one is merged. This should be clearly reflected in the tests when they run.

SamTov:
One conflict between the espresso and resobee engine is the direct use of numpy arrays as actions instead of classes. If you can wrap your actions into classes, it would make the two engines coexist comfortably.

CI/resobee_tests/resobee_example.ipynb
SamTov, line 1: please move this to the SwarmRL examples repository with a new directory for Resobee tutorials. Also, make sure all of the outputs are deleted before pushing so as to avoid blowing up the size of the files.

CI/resobee_tests/resobee_example.py
SamTov, line 1: If this is to be a real test, it should be under `integration_tests` and have some kind of output check to be sure that it has worked correctly.

CI/resobee_tests/resobee_example.py
```diff
@@ -0,0 +1,119 @@
...
+
+import flax.linen as nn
+import numpy as np
+import optax
+import yaml
+ 
+import swarmrl as srl
+from swarmrl.actions.actions import Action
+
+# %% [markdown]
```
SamTov, line 14: Please remove commented code.

requirements.txt
```diff
@@ -12,6 +12,8 @@ setuptools~=58.0.4
 optax
 rich
 dataclasses~=0.6
+pyzmq  # requires libzmq3-dev to work
```
SamTov, line 15: Should we ship SwarmRL with resobee as a default, or do we add it as an optional package?

swarmrl/agents/actor_critic.py
```diff
@@ -172,7 +172,7 @@ def calc_action(self, colloids: typing.List[Colloid]) -> typing.List[Action]:
         action_indices, log_probs = self.network.compute_action(
             observables=state_description
         )
-        chosen_actions = np.take(list(self.actions.values()), action_indices, axis=-1)
+        chosen_actions = np.take(list(self.actions.values()), action_indices, axis=0)#for Resobee "0", for Espresso "-1"
```
SamTov, line 175: This is where, for example, if you moved the resobee actions were classes, the code would be compatible.

swarmrl/engine/resobee.py
```diff
@@ -0,0 +1,156 @@
+"""
```
SamTov, line 1: Can you double check that we have doc strings and type hints in all methods here.

swarmrl/observables/concentration_field.py
```diff
@@ -124,15 +124,20 @@ def compute_observable(self, colloids: List[Colloid]):
...
-            msg = (
-                f"{type(self).__name__} requires initialization. Please set the "
-                "initialize attribute of the gym to true and try again."
-            )
-            raise ValueError(msg)
-
-        observables = [
-            self.compute_single_observable(index, colloids) for index in reference_ids
-        ]
+        # if self._historic_positions == {}:
```
SamTov, line 127: Can you delete the commented out code.

swarmrl/tasks/object_movement/resobee_swarm.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+
+    def __init__(
+        self,
+        box_length: np.ndarray,
+        particle_type: int = 0,
+        repulsion_radius: float=1,
+        alignment_radius: float=1,
+        desired_speed:float =1,
+
```
SamTov, line 28: Please remove the spaces.

swarmrl/tasks/object_movement/resobee_swarm.py
```diff
@@ -0,0 +1,116 @@
...
+        chosen_colloids = [
+            colloid for colloid in colloids if colloid.type == self.particle_type
+        ]
+        colloid_positions = np.array([colloid.pos for colloid in chosen_colloids])
+        colloid_velocities = np.array([colloid.velocity for colloid in chosen_colloids])
+        alignment = lambda v_i, v_j: np.dot(v_i,v_j)
+        repulsion = lambda dist: dist/24.0-1/24.0
+        reward_homing = -np.linalg.norm(colloid_positions/self.box_length-0.5,axis = -1)
+        # reward_alignment = 0
```
SamTov, line 111: Commented code here can go.

swarmrl/tasks/object_movement/resobee_swarm.py
SamTov, line 1: In general, I would question whether this needs to be a part of the code right now. I would like to move to a system where only tasks with an associated doi are included in the package to have some real reason for them being there. Perhaps move it to the example notebook as a custom class.

swarmrl/trainers/continuous_trainer.py
```diff
@@ -11,7 +11,7 @@
 
 class ContinuousTrainer(Trainer):
     """
-    Class for the simple MLP RL implementation.
+    Class for the simple MLP RL implementation. Behaves like epsiodic trainer for ResoBee.
```
SamTov, line 14: Why not just use the episodic trainer then? This just should never be used with ResoBee in this case, or ResoBee should be adapted to handle continuous training.

swarmrl/trainers/continuous_trainer.py
```diff
@@ -47,8 +47,9 @@ def perform_rl_training(
         force_fn = self.initialize_training()
 
         # Initialize the tasks and observables.
-        for agent in self.agents.values():
-            agent.reset_agent(self.engine.colloids)
+        
+        # for agent in self.agents.values():
```
SamTov, line 51: Please remove the commented code.

swarmrl/trainers/episodic_trainer.py
```diff
@@ -1,21 +1,17 @@
 """
-Module for the EpisodicTrainer
+Module to implement a simple multi-layer perceptron for the colloids.
```
SamTov, line 2: Why was this changed?

swarmrl/trainers/episodic_trainer.py
SamTov, line 1: Can you explain why this module is so heavily changed?

SamTov (changes requested):
I have a few comments so far, although I have not done a thorough review yet. We should work out if Resobee should be optional, which I am leaning towards. Please go through what was written and try to address each point. It also seems like tests aren't running but I am not sure why. This also means the fomatting has not run which will clean up parts of the code.

swarmrl/agents/actor_critic.py
Mag-r: ResoBee now also uses the action class.

swarmrl/engine/resobee.py
Mag-r: Added doc strings and type hints everywhere.

swarmrl/trainers/continuous_trainer.py
Mag-r: removed the commented code.

swarmrl/observables/concentration_field.py
Mag-r: deleted the commented code.

swarmrl/trainers/episodic_trainer.py
Mag-r: ResoBee doesn't need to initialize its observables and everything Espresso needs to reset it's simulation state, ResoBee does it automatically. 
I'd say that this is the major conflict between ResoBee and Espresso.

Mag-r: moved changes from episodic_trainer to resobee. ResoBee and Espresso now use the same episodic trainer.

swarmrl/trainers/continuous_trainer.py
Mag-r: removed the comment.

CI/resobee_tests/resobee_example.ipynb
Mag-r: moved to example repo.

CI/resobee_tests/resobee_example.py
Mag-r: it is the same as the jupyter notebook. Removed it.

requirements.txt
Mag-r: I think optional would be better.

swarmrl/tasks/object_movement/resobee_swarm.py
Mag-r: removed the task.

SamTov:
I think we should shelve this for the time being. The branch can stay open but I don't see it being merged in the near future.

PR #83, opened 2024-02-28
Make ZnNL import optional

SamTov:
Wrap the ZnNL import in a try except to avoid forcing the import.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,5 @@
 [flake8]
-ignore = E226,E302,E41, W503, E203
+ignore = E226,E302,E41, W503, E203, E402
 max-line-length = 88
 exclude =
     .git,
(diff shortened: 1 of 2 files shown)


.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-ignore = E226,E302,E41, W503, E203
+ignore = E226,E302,E41, W503, E203, E402
```
christophlohrmann, line 2: what is that?

swarmrl/intrinsic_reward/__init__.py
```diff
@@ -2,13 +2,29 @@
...
+        RNDConfig.__name__,
+        RNDReward.__name__,
+    ]
+except ImportError:
+    logger.info(
+        "Could not find optional packages, certain features might be unavailable."
+    )
+    pass
 
 __all__ = [
```
christophlohrmann, line 28: This will overwrite the ``__all__`` above

SamTov: Fixed it.

christophlohrmann (approved):

PR #84, opened 2024-03-05
semi episodic file writing fixed.

Dalouvid:
Fix #

* the trajectories of the simulations are now all saved to the same file 
* for each cycle (all episodes between two resets) there is a new group in the hdf5 database
* for none episodic or the case that one does not look to that data, the group is still called "colloids" 
* 

```

```

### DIFF
--- a/CI/espresso_tests/integration_tests/test_rl_trainers.py
+++ b/CI/espresso_tests/integration_tests/test_rl_trainers.py
@@ -244,6 +244,7 @@ class EspressoTestRLTrainers(ut.TestCase):
                 n_episodes=10,
                 system=self.system,
                 episode_length=10,
+                save_episodic_data=False,
             )
 
             dumped_files = glob.glob(f"{temp_dir}/episodic/*")
@@ -300,6 +301,7 @@ class EspressoTestRLTrainers(ut.TestCase):
                 reset_frequency=1000,  # Will only be reset after a failure.
                 system=self.system,
                 episode_length=10,
+                save_episodic_data=False,
             )
 
             dumped_files = glob.glob(f"{temp_dir}/episodic/*")
@@ -348,6 +350,7 @@ class EspressoTestRLTrainers(ut.TestCase):
                 system=self.system,
                 reset_frequency=2,
                 episode_length=10,
+                save_episodic_data=False,
             )
 
             dumped_files = glob.glob(f"{temp_dir}/episodic/*")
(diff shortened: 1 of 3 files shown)


CI/espresso_tests/integration_tests/test_rl_trainers.py
```diff
@@ -352,6 +353,63 @@ def get_engine(system):
...
+            )
+
+            with h5py.File(
+                f"{temp_dir}/episodic/data_writing/trajectory.hdf5", "r"
+            ) as file:
+                keys = list(file.keys())
+
+            assert len(keys) == 10
+            assert keys == ["0", "1", "2", "3", "4", "5", "6", "7", "8", "9"]
+            dumped_files = glob.glob(f"{temp_dir}/episodic/*")
```
SamTov, line 410: It would be nice to add a small ds size check here as well, just so we know that the correct data is being stored in each one.

swarmrl/trainers/episodic_trainer.py
```diff
@@ -49,6 +50,9 @@ def perform_rl_training(
                 After how many episodes is the simulation reset.
         load_bar : bool (default=True)
                 If true, show a progress bar.
+        save_episodic_data : bool (default=False)
```
SamTov, line 53: That's a really cool thing to have, I would only suggest that it be set default to `True`

swarmrl/trainers/episodic_trainer.py
```diff
@@ -79,9 +83,22 @@ def perform_rl_training(
             )
             for episode in range(n_episodes):
                 # Check if the system should be reset.
-                if episode % reset_frequency == 0 or killed:
+                if episode % reset_frequency == 0 and reset_frequency > 0 or killed:
```
SamTov, line 86: Can it be true that `reset_frequency == 0` but `episode % reset_frequency == 0`? I don't know for sure, I haven't tested it, I would just be surprised if the remainder of infinity could be defined. In general, what issue prompted this addition to the loop condition?

SamTov (changes requested):
That's a really helpful PR! I am very happy with it, thank you. I have a couple of questions and comments but if you address each of them, I will approve.

swarmrl/trainers/episodic_trainer.py
Dalouvid: I set it to `True` and changed the other tests accordingly.

swarmrl/trainers/episodic_trainer.py
Dalouvid: True, that's nonsense. 
Changed it such that it checks if `episode == 0 `.

SamTov (approved):

PR #87, opened 2024-04-24
added logging to proximal_policy_loss.py

jdrotleff:
added logging feature to proximal_policy_loss.py as in policy_gradient_loss.py 

## Summary by CodeRabbit

* **Chores**
  * Added enhanced logging infrastructure to improve debugging capabilities during training operations with better output capturing and intermediate value tracking.


SamTov (changes requested):
Thanks for the PR! Can you shift the LoggerWriter out of this module and into the utils? Further, just to clarify, this will only store everything if we set the mode to debug, write?

PR #91, opened 2024-05-13
Add Gaurav's simulations as an engine

christophlohrmann:
Done:

- Basic implementation of eq of motion and integration
- Trajectory writing
- Test if it runs

To Do:

- Units
- Actual testing
- Tying in the new set of actions into the framework of our existing actions
- Reproducing some simulations from the paper

Future:

- Implement RL models that use the new actions


swarmrl/engine/gaurav_sim.py
```diff
@@ -0,0 +1,697 @@
...
+    if x == 0.0:
+        return 0.0212764 / radius
+    else:
+        logx = np.log(x)
+        return ((0.0212764 * (-logx) + 0.157378) * (-logx) + 0.269886) / (
+            radius * (-logx) * ((-logx) + 6.0425) + 6.32549
+        )
+
+
+@numba.jit
```
SamTov, line 184: Is there a reason to with numba here and not Jax? If one requirement already includes GPU support and JIT compile I would lean towards sticking with that unless there is a reason not to.

swarmrl/engine/gaurav_sim.py
```diff
@@ -0,0 +1,697 @@
...
+                r_ij_vec = r_ijs[j, :]
+                r_ij_norm = r_ij_norms[j]
+                r_ij_angle = np.arctan2(r_ij_vec[1], r_ij_vec[0])
+                r_ij_unit_vec = r_ij_vec / r_ij_norm
+                r_ij_cross_z = np.array([r_ij_unit_vec[1], -r_ij_unit_vec[0]])
+                edge_edge_dist = edge_edge_dists[j]
+                phi_i = state[i, 2] - r_ij_angle
+                phi_ij = phi_i
+
+                # f_mag_on_prefactor = (
```
SamTov, line 569: Do we need the old code here?

swarmrl/engine/gaurav_sim.py
```diff
@@ -0,0 +1,700 @@
...
+import shelve
+import typing
+
+import h5py
+import numba
+import numpy as np
+import pint
+import scipy.integrate
+
+from swarmrl.engine import Engine
```
SamTov, line 12: I added the full import path.

swarmrl/engine/gaurav_sim.py
```diff
@@ -0,0 +1,700 @@
...
+            return self.get_rhs(t, state)
+
+        state_flat = self._get_state_from_rafts().flatten()
+
+        n_snapshots_per_slice = int(
+            round(self.params.time_slice / self.params.snapshot_interval)
+        )
+        for _ in range(n_slices):
+            # Check for a model termination condition.
+            if model.kill_switch:
```
SamTov: I added a `kill_switch` check.

swarmrl/engine/gaurav_sim.py
```diff
@@ -0,0 +1,700 @@
...
+
+@dataclasses.dataclass
+class GauravAction:
+    amplitudes: np.array
+    frequencies: np.array
+    phases: np.array
+    offsets: np.array
+
+
+class Model:
```
SamTov, line 104: Is it possible to have these inherit from our `ForceFunction`? It sees like most of the code can be reused we should just rename `colloids` to `particles` or something along those lines. Perhaps we make a more general parent for the force functions so that we can have an ESPResSo and newer ones which then get their own custom type hints and we can do some checks when users try to use espresso actions in an MPI simulation. What do you think?

swarmrl/engine/gaurav_sim.py
```diff
@@ -0,0 +1,700 @@
...
+            # small wrapper to remove it from the call signature
+            return self.get_rhs(t, state)
+
+        state_flat = self._get_state_from_rafts().flatten()
+
+        n_snapshots_per_slice = int(
+            round(self.params.time_slice / self.params.snapshot_interval)
+        )
+        for _ in range(n_slices):
+            # Check for a model termination condition.
```
SamTov, line 674: I added a kill switch here but commented it out as without the main parent classes for the force model it will break the tests.

SamTov (changes requested):
It looks good to me; I don't have a good understanding of the simulation side of things. Depending on how broad it will be, we could update the documentation side, but I would leave this for Gaurav. 

I had a few comments regarding old code and the use of numba. I would prefer to just use Jax as it will also then compile nicely with the other parts of SwarmRL which also compile with Jax. I think it will also reduce the number of large dependencies. But if there is a real reason for it then I am open. 

The other larger point is the use of a new `Model`. I think it would also be possible to reuse our `ForceFunction` parent class or, as I write in the comment, make a slightly more general parent class and start branching off with Espresso and MPI children.

PR #94, opened 2024-06-04
Update espresso version

phohenberger:
Update the espresso version to the current commit


christophlohrmann (approved):
if the tests pass, I'm always in favour of keeping our espresso up-to-date

PR #96, opened 2024-09-23
Checkpointer

jdrotleff:
# Checkpointer

A class that can be used to automatically save the current Neural Network if the reward or its running average meet some chosen criteria.
To use it, pass a checkpoint dictionary to the Trainer, in which you define your wanted save-criteria.

Currently, these options are implemented.

- Regular Backup: Saves a copy of the network every ```n``` episodes.
- Goal Backup: Saves a copy of the network as soon as a desired reward threshold is
reached. The simulation can possibly be stopped after the goal is reached, continued for a defined # of episodes or until it would end normally. 
- Best Model: Always saves a copy of the network, if the mean reward of the last ```window_width```
episodes is greater than before. The next best reward can be increased as a percentage. 
- Backup Model: Saves a copy of the network, if the mean reward of the last ```window_width``` episodes
stops increasing. This could possibly save models, that still did some learning while the
agents were successful. Alternatively, the backup model can be compared to the best
model to analyze the phenomenom of forgetting.

For every episode each criterion is checked. If at least one is fulfilled, the model gets exported. The name of the exported model contains the criteria that are met.

An example application of this class can look like this:
```
checkpointing_params = {'window_width' : 10,
                        'DO_GOAL_MODEL' : True,
                        'required_reward' : 250,
                        'DO_GOAL_BREAK' : True,
                        'running_out_length' : 25,
                        'DO_BEST_MODEL' : True,
                        'min_reward' : 250,
                        'increase_factor' : 1.05,
                        'DO_BACKUP_MODEL' : True,
                        'min_backup_reward': 250,
                        'DO_REGULAR' : True,
                        'save_models_intervall' : 25,
                        }

rl_trainer = srl.trainers.EpisodicTrainer(
    [agent], 
    checkpointing_params
)
rewards = rl_trainer.perform_rl_training(...)
```
The chosen parameters depend highly on the simulation performed.

For a more detailled explanation check the docstrings in trainer.py.

## Summary by CodeRabbit

- **New Features**
  - Built-in training checkpointing with multiple strategies (regular interval, best-reward, goal-based, backup).
  - Trainers export checkpoints during episodes and can early-stop when a reward goal is reached.
  - Trainer accepts configurable checkpointers and auto-selects checkpoint path.
  - Checkpointers exposed as a public module for easy access.

- **Refactor**
  - Episode rewards stored as a fixed-size array; added logging hooks and per-episode checkpoint integration.

- **Tests**
  - Added unit tests covering all checkpointer strategies, initialization, and trigger/stop behavior.


SamTov:
I am more a fan of having a generic base class for checkpointing and then having child classes for the different kinds of checkpointing. This avoids having huge parameter dictionaries. We can then allow users to pass multiple checkpointers in a list.

jpd-de:
I changed the structure of the checkpointers to a base class/child class structure.


To use it, create Checkpointers of your choice and pass them to the Trainer.

Currently, these options are implemented.

- RegularCheckpointer: Saves a copy of the network every ```save_interval``` episodes.
- GoalCheckpointer: Saves a copy of the network as soon as a desired reward threshold is
reached. The simulation can possibly be stopped after the goal is reached, continued for a defined # of episodes or until it would end normally. 
- BestRewardCheckpointer: Always saves a copy of the network, if the mean reward of the last ```window_width```
episodes is greater than before. The next best reward can be increased as a percentage. 
- BackupCheckpointer: Saves a copy of the network, if the mean reward of the last ```window_width``` episodes
stops increasing. This could possibly save models, that still did some learning while the
agents were successful. Alternatively, the backup model can be compared to the best
model to analyze the phenomenom of forgetting.

General parameters:
```window_width```: How many episodes should be considered for the running reward average?
GoalCheckpointer parameters:
```running_out_length```: How many episodes should the simulation continue, as the goal is reached.
```required_reward```: Sets the wanted reward.
BestRewardCheckpointer and BackupCheckpointer parameters.
```increase_factor```: The factor the reward should be greater than the previous maximum reward.
```wait_time```: Sets the # of episodes that will not be saved after a best model was found. This prevents a huge amount of saves of the same model.


For every episode each criterion is checked. If at least one is fulfilled, the model gets exported. The name string of the exported model contains the criteria that are met.

An example use of all implemented methods could look like this:
```
from swarmrl.checkpointers import GoalCheckpointer, BestRewardCheckpointer, BackupCheckpointer, RegularCheckpointer

Goal_Checkpointer = GoalCheckpointer(required_reward=250, 
                                            running_out_length=25, 
                                            window_width=10,
                                            do_goal_break=True)
Best_Checkpointer = BestRewardCheckpointer(min_reward=250,
                                            increase_factor=1.05,
                                            window_width=10)
Backup_Checkpointer = BackupCheckpointer(min_backup_reward=250,
                                            wait_time=10,
                                            window_width=10)
Regular_Checkpointer = RegularCheckpointer(save_interval=25)

rl_trainer = srl.trainers.EpisodicTrainer(
    [agent], 
    [Goal_Checkpointer, Backup_Checkpointer, Regular_Checkpointer]
    )
```
Note that the chosen parameters depend highly on the simulation performed and your specific task.

File: swarmrl/checkpointers/__init__.py [hunk 1]
```diff
@@ -0,0 +1,17 @@
+"""
+Package to hold all SwarmRL Checkpointers.
+"""
+
+from swarmrl.checkpointers.backup_checkpointer import BackupCheckpointer
+from swarmrl.checkpointers.best_checkpointer import BestRewardCheckpointer
+from swarmrl.checkpointers.checkpointer import Checkpointer
```
SamTov, line 7: This will seem pedantic but can we change `Checkpointer` to `BaseCheckpointer` to be very clear that it is a parent class.

swarmrl/checkpointers/backup_checkpointer.py
```diff
@@ -0,0 +1,54 @@
...
+import numpy as np
+
+from swarmrl.checkpointers.checkpointer import Checkpointer
+
+logger = logging.getLogger(__name__)
+
+
+class BackupCheckpointer(Checkpointer):
+    """
+    Checkpointer that saves the model if a new best reward is achieved.
```
SamTov, line 16: Is this correct?

swarmrl/checkpointers/best_checkpointer.py
```diff
@@ -0,0 +1,64 @@
...
+logger = logging.getLogger(__name__)
+
+
+class BestRewardCheckpointer(Checkpointer):
+    """
+    Checkpointer that saves the model if a new best reward is achieved.
+    """
+
+    def __init__(
+        self, min_reward=250, increase_factor=1.05, window_width=30, wait_time=10
```
SamTov, line 20: Can you add some doc string for this function to introduce each argument?

swarmrl/trainers/episodic_trainer.py
```diff
@@ -65,7 +69,7 @@ def perform_rl_training(
         simulation, the system will be reset.
         """
         killed = False
-        rewards = [0.0]
+        rewards = np.zeros(n_episodes)
```
SamTov, line 72: Nice addition.

SamTov (changes requested):
In general, I like it. I would like to see some more comprehensive documentation as well as some changes to the structure of a few checkpointers and names. Please take a look at the comments.

swarmrl/checkpointers/__init__.py
jpd-de: Sure! Makes it clearer.

swarmrl/checkpointers/best_checkpointer.py
jpd-de: Done.

swarmrl/trainers/episodic_trainer.py
jpd-de: Thanks!

swarmrl/checkpointers/backup_checkpointer.py
```diff
@@ -0,0 +1,101 @@
+"""
+Module for the Save-a-Backup-Checkpointer
+"""
+
+import numpy as np
+
+from swarmrl.checkpointers.base_checkpointer import BaseCheckpointer
+
+
+class BackupCheckpointer(BaseCheckpointer):
```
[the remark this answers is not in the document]
jdrotleff: The idea was to save a checkpoint when forgetting occurs. So it saves a checkpoint if the performance drops after a reward peak, but is still above some minimum threshold (to not save a lot of noise at low rewards). This threshold is updated to the current average reward when a checkpoint is stored, so that the next checkpoint made should be a model with a better reward.
We could rename it for clarity.
```suggestion
class OnDeclineCheckpointer(BaseCheckpointer):
```
Alternatively, see the answer to your next comment

swarmrl/checkpointers/best_checkpointer.py
```diff
@@ -0,0 +1,86 @@
+"""
+Module for the Save-Best-Checkpointer
+"""
+
+import numpy as np
+
+from swarmrl.checkpointers.base_checkpointer import BaseCheckpointer
+
+
+class BestRewardCheckpointer(BaseCheckpointer):
```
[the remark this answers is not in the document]
jdrotleff: Performance is different, yes.
We can remove the `BackupCheckpointer`and add a flag `only_save_on_decline` here. 

Then we might need to think about keeping a copy of the previous episode's model, as it was performing better than the current one, if reward is dropping again, or do we?

swarmrl/checkpointers/base_checkpointer.py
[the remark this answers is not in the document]
jdrotleff: Currently in the trainer. I saw your comment there. Thanks for pointing it out! I'll see how it can be done

swarmrl/trainers/continuous_trainer.py
```diff
@@ -64,26 +69,49 @@ def perform_rl_training(
...
+                    )
                     system_runner.finalize()
                     break
 
-                rewards.append(current_reward)
-                episode += 1
+                if len(self.checkpointers) > 0:
+                    export = []
+                    save_string = ""
+                    for checkpointer in self.checkpointers:
```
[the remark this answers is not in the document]
jdrotleff: I think we should add a `CheckpointManager` class that uses a callback to the export method of the trainer, manages the `save_string`logic and all active checkpointers.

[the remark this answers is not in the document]
SamTov: What about the parent class?

swarmrl/checkpointers/best_checkpointer.py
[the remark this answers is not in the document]
SamTov: After discussion, I am happy with this.

swarmrl/trainers/continuous_trainer.py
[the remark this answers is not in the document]
jdrotleff: Addressed it

swarmrl/trainers/trainer.py
```diff
@@ -115,9 +151,54 @@ def export_models(self, directory: str = "Models"):
...
+        """
+        if self.checkpoint_manager is None:
+            return False
+        return self.checkpoint_manager.check_and_save(
+            rewards=rewards,
+            current_episode=episode,
+            current_reward=current_reward,
+        )
+
+    def maybe_stop_training(self) -> tuple[bool, int]:
```
SamTov: I do love this name. But perhaps we can find something a bit more technical like `check_for_stop_criterion`?

SamTov (approved):
LGTM

PR #97, opened 2024-09-26
Use a standard test driver for the ESPResSo testsuite

### LINKED ISSUE #85: Avoid directly calling the shell from Python scripts
**Describe the bug**
Some scripts use [`os.system()`](https://docs.python.org/3/library/os.html#os.system) to create a subshell. This is discouraged by the Python standard. See pyMBE-dev/pyMBE#16 for the dangers and pitfalls of `os.system()`, and suggestions for alternatives that are both safer and more interoperable.

Here is an example:
https://github.com/SwarmRL/SwarmRL/blob/fb0f3c4f00201f226cb9b62f569b51a1dc1b9077/CI/run_espresso_test_suite.py#L21-L23

There are several bugs:
* at line 21: the subshell command assumes the file path either doesn't contain characters with special meaning in shell (whitespace, quote, bracket, etc.) or has them properly escaped
* at line 21: the `SIGINT` signal (triggered by Ctrl+C on Linux) is silently ignored by the main process: the currently running subshell will interrupt and return a non-zero error code, but the Python script will resume execution and continue spawning new subshells in a loop, which the user has to interrupt manually by sending multiple signals
* at line 22: the command assumes that non-zero error codes mean a failure, which is incorrect: Python 3.12 introduced error code 5 for when [`unittest.main()`](https://docs.python.org/3/library/unittest.html#unittest.main) attempted to run tests but had to skip all of them due to missing dependencies or any other valid reason, and therefore the testsuite result is neither a success nor a failure

The last point is quite problematic, because Python 3.12 will become available in Ubuntu 24.04 soon, with a release date set to April 25, 2024. This change in the API of Python 3.12 was communicated to ESPResSo users in January 2024 in multiple venues (see e.g. [ESPResSo meeting notes](https://github.com/espressomd/espresso/wiki/Proceedings-2024-ESPResSo-meetings#2024-01-30) and [ICP Mattermost](https://chat.icp.uni-stuttgart.de/icp/pl/w61amwgts7rexfhzqotmwqs6wo)). Also one cannot check if `exit_cmd` is equal to 5, because the value returned by `os.system()` is OS-dependent. For example on Ubuntu, a bitmask operation is needed to extract the relevant byte from the concatenated tuple (see [`os.wait()`](https://docs.python.org/3/library/os.html#os.wait) documentation). A perfectly equivalent and cross-platform alternative can be expressed with [`subprocess.run()`](https://docs.python.org/3/library/subprocess.html#subprocess.run), which takes care of escaping file paths and wrapping the error code in a data member.

**To Reproduce**
Steps to reproduce the behavior from the interrupt signal:
1. Go to the top-level directory of the SwarmRL repository
2. Run the CI command `espresso/build/pypresso CI/run_espresso_test_suite.py`
3. Wait for the first test to display a progress bar
4. Try to stop the testsuite execution with SIGINT (Ctrl+C on Linux)

Steps to reproduce the behavior from the API change:
1. Set up an environment with Python 3.12
2. Create a test file `test.py` containing the following code:
   ```py
   import unittest as ut
   
   @ut.skip("reason")
   class Test(ut.TestCase):
       pass
   
   if __name__ == "__main__":
       ut.main()
   ```
4. Open a Python 3.12 terminal in the same folder as `test.py`
5. Call the test with `os.system()`:
   ```py
   import os
   err = os.system("python3 test.py")
   print(err)
   ```
6. See error code 1280 (depending on the OS and runtime environment)
7. Call the test with `subprocess.run()`:
   ```py
   import subprocess
   out = subprocess.run(["python3", "test.py"])
   print(out.returncode)
   ```
8. See error code 5

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04

jngrad:
Fix #85

* document how to contribute to SwarmRL: installing dependencies, running tests, running Sphinx
* rewrite CI workflows to address GitHub Actions warnings and YAML syntax errors
* replace legacy test driver by CTest, with support for code coverage and multi-threading
   * more details in https://github.com/pyMBE-dev/pyMBE/pull/87#issuecomment-2316278053.

```sh
export PYTHONPATH="${HOME}/Documents/espresso/build/src/python"
sh CI/run_espresso_test_suite.sh -j $(nproc)
```


SamTov (approved):
I feel like you should be on the paper after that contribution... Thanks for cleaning things up!

PR #102, opened 2025-04-14
Shift reward

Mag-r:
Before the state and the reward were calculated at the same time. But in general the reward depends on the action and the next state. This means the reward needed to be shifted inside the loss calculation. This was not done. 
With this change the tuple (s, a, r, s') can all be accessed with the same index. 

The external reward argument isnt needed for most cases, but it is required when using a gym environment.

## Summary by CodeRabbit

* **New Features**
  * Agents now support explicit reward computation combining environment task rewards with optional external rewards.

* **Refactor**
  * Reward computation is now separated from action selection logic for improved modularity and cleaner control flow in agent decision-making.


jdrotleff:
The latest PR #119 was built on this PR, so we can close it now.

PR #105, opened 2025-10-13
Remove RND as hotfix for tests

SamTov:
Remove ZnNL and RND from our losses to have a hotfix for the failing tests and build.

## Summary by CodeRabbit

* **Chores**
  * Simplified CI installation by switching to a standard editable install; test flow unchanged.
  * Removed a previously available optional install extra, so that related optional dependencies are no longer installed by default.
  * Updated an internal submodule reference with no functional impact.

* **Tests**
  * Minor test adjustment to use a different tree traversal utility for state comparisons; behavior unchanged.


jpd-de (approved):
LGTM

PR #112, opened 2026-03-25
Disable logging of JAX values by default

jdrotleff:
Users need to manually turn it on if they want to log JAX values


SamTov (approved):
Thanks for the quick action

PR #113, opened 2026-03-27
Refactor HDF5 trajectory storage, add agent data saving

jdrotleff:

- moved trajectory writing into shared storage utilities so agent and simulation data use the same HDF5 storage path.
- `espresso.py` engine no longer writes HDF5 inline; it now uses SimulationTrajectoryStorage.
- Agent trajectory saving is configurable via `AgentTrajectoryStorage` and `AgentStorageConfig`, either by preset, or by explicitly selecting attributes  (`actions`, `rewards`, `log_probs`, `features`, `killed`).
- Agent trajectory writes can be batched with `write_chunk_size`; buffered data is written on `finalize()`.
- Trajectory outputs are safe by default: writing fails if the target file already exists unless existing-file reuse is explicitly allowed.
-  Existing HDF5 files can be reused intentionally, for example to store multiple simulation cycles in separate HDF5 groups during semi-episodic training.
- Extends test coverage for file safety, attribute validation, appending new data, vector feature persistence, and correct behavior when no new samples are available, thus no write is performed.

Parts of this PR were drafted with AI assistance, changes were reviewed and validated manually.

#### Examples

1. Store minimally

```
from swarmrl.agents.actor_critic import ActorCriticAgent
from swarmrl.utils.storage_utils import AgentStorageConfig

storage_cfg = AgentStorageConfig(
    out_folder="runs/exp_minimal",
    storage_preset="minimal",       # default behavior
    allow_existing_file=False,      # default safety
)

agent = ActorCriticAgent(
    particle_type=0,
    network=network,
    observable=observable,
    actions=actions,
    task=task,
    storage_config=storage_cfg,
)
```

2. Store all supported fields
```
storage_cfg = AgentStorageConfig(
    out_folder="runs/exp_all",
    storage_preset="all",
    allow_existing_file=False,
)
```
3. Explicit attributes to store

```
storage_cfg = AgentStorageConfig(
    out_folder="runs/exp_custom",
    stored_attributes=["actions", "rewards", "killed"],
    allow_existing_file=False,
)
```
4. Batch agent trajectory writes

```
storage_cfg = AgentStorageConfig(
    out_folder="runs/exp_batched",
    storage_preset="all",
    write_chunk_size=10,
    allow_existing_file=False,
)
```
5. Reuse an existing HDF5 file intentionally
```
storage_cfg = AgentStorageConfig(
    out_folder="runs/exp_reuse",
    storage_preset="minimal",
    allow_existing_file=True,
)
```

### DIFF
--- a/CI/espresso_tests/integration_tests/test_rl_trainers.py
+++ b/CI/espresso_tests/integration_tests/test_rl_trainers.py
@@ -379,6 +379,7 @@ class EspressoTestRLTrainers(ut.TestCase):
                     system=self.system,
                     write_chunk_size=10,
                     h5_group_tag=cycle_index,
+                    fail_if_trajectory_file_exists=False,
                 )
 
                 coll_type = 0
--- a/CI/espresso_tests/integration_tests/test_rl_trainers.py
+++ b/CI/espresso_tests/integration_tests/test_rl_trainers.py
@@ -51,7 +51,7 @@ class KillTask(srl.tasks.Task):
     Dummy task for the tests.
 
     This task will turn on the killswitch after
-    4 episodes.
+    15 episodes.
     """
 
     i: int = 0
--- a/CI/espresso_tests/integration_tests/test_rl_trainers.py
+++ b/CI/espresso_tests/integration_tests/test_rl_trainers.py
@@ -428,7 +428,7 @@ class EspressoTestRLTrainers(ut.TestCase):
                     system=self.system,
                     write_chunk_size=10,
                     h5_group_tag=cycle_index,
-                    fail_if_trajectory_file_exists=False,
+                    allow_existing_trajectory_file=True,
                 )
 
                 coll_type = 0
--- a/CI/espresso_tests/integration_tests/test_rl_trainers.py
+++ b/CI/espresso_tests/integration_tests/test_rl_trainers.py
@@ -51,7 +51,7 @@ class KillTask(srl.tasks.Task):
     Dummy task for the tests.
 
     This task will turn on the killswitch after
-    15 episodes.
+    4 episodes.
     """
 
     i: int = 0
@@ -181,7 +181,7 @@ class EspressoTestRLTrainers(ut.TestCase):
                 self.ureg.Quantity(400, "micrometer"),
                 type_colloid=0,
             )
-            # We need a custom protoc0l for this test.
+            # We need a custom protocol for this test.
             agent = srl.agents.ActorCriticAgent(
                 particle_type=0,
                 network=self.network,
@@ -281,7 +281,7 @@ class EspressoTestRLTrainers(ut.TestCase):
 
                 return system_runner
 
-            # We need a custom protoc0l for this test.
+            # We need a custom protocol for this test.
             agent = srl.agents.ActorCriticAgent(
                 particle_type=0,
                 network=self.network,
--- a/swarmrl/utils/storage_utils/trajectory_storage.py
+++ b/swarmrl/utils/storage_utils/trajectory_storage.py
@@ -24,8 +24,13 @@ class DictTrajectoryStorage(HDF5TrajectoryStorage):
         h5_group_tag: str,
         dataset_specs_builder: Callable[[Any], Dict[str, Dict[str, Any]]],
         sample_extractor: Callable[[Any], Dict[str, Any]],
+        fail_if_exists: bool = False,
     ):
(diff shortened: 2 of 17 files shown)


jdrotleff:
Quick note, I don't really understand why the tests wasn't passing for `a0edde1`.

Tests were passing for the last actual HDF5 related commit`b6ca794`.
Tests were also passing for [PR 114](https://github.com/SwarmRL/SwarmRL/pull/114).

Only when merging master with merged PR 114 into this branch, the errors occured.

Please have a look at the CI change I've made.

CI/espresso_tests/integration_tests/test_rl_trainers.py
```diff
@@ -51,7 +51,7 @@ class KillTask(srl.tasks.Task):
     Dummy task for the tests.
 
     This task will turn on the killswitch after
-    4 episodes.
+    15 episodes.
```
SamTov: Why go longer here?

swarmrl/utils/storage_utils/trajectory_storage.py
```diff
@@ -0,0 +1,289 @@
...
+
+from dataclasses import dataclass
+from typing import Any, Callable, Dict, List
+
+import numpy as np
+
+from swarmrl.utils.storage_utils.core_storage import HDF5TrajectoryStorage
+
+
+class DictTrajectoryStorage(HDF5TrajectoryStorage):
```
SamTov: Can you explain this one a bit. I don't understand what this data structure looks like and why a dictionary would be a child of HDF5.

SamTov (changes requested):
So I have two comments in the actual PR, otherwise it looks good. 

One small thing, what is the update procedure during a simulation? Do you open and close everything on every tilmestep? Is there a dump parameter.

CI/espresso_tests/integration_tests/test_rl_trainers.py
jdrotleff: The kill switch is implemented for 15 episodes, I just synced the doc string with the code.

swarmrl/utils/storage_utils/trajectory_storage.py
jdrotleff: There is no actual dictionary being a child of HDF5 here. The name `DictTrajectoryStorage` was badly chosen.

The class was intended as an adapter layer. It should convert different Python trajectory objects into dictionaries of the form `{dataset_name: value}` where each key corresponds to an HDF5 dataset name.
For example, simulation data would become keys like `Times`, `Ids`, and `Velocities`, while agent data would become keys like `actions` and `rewards`.

So the HDF5 structure is still only groups and datasets. The dictionary was just an intermediate Python representation used before writing to HDF5.

After looking at it again, I share your doubt about this class. The shared HDF5 base class can own the generic buffering/writing logic, while `AgentTrajectoryStorage` and `SimulationTrajectoryStorage` can directly provide their dataset specs and extracted samples. 

I'll remove `DictTrajectoryStorage`and flatten the hierarchy accordingly.

jdrotleff:
Concerning your other questions
* **What is the update procedure during simulation, do you open and close everything on every tilmestep?**

The HDF5 file is not opened and closed on every MD timestep. 
For simulation trajectories, the PR only moved the actual HDF5 append logic into `SimulationTrajectoryStorage`. The existing buffering behavior in `espresso.py` stayed the same, that is:
1. `MDParams.write_interval` controls how often simulation data is sampled and added to the trajectory buffer.
2. `EspressoMD.write_chunk_size` sets how many frames are collected before being flushed to the HDF5 file.

For agent data, it is stored, if activated, on every `update_agent()` step (only for the actor critic agent at the moment).

* **Is there a dump parameter?**
For simulation trajectories, `write_chunk_size` and `write_interval` can be used for this. 
For agent there is not yet one, I'll add one, good point.

CI/espresso_tests/integration_tests/test_rl_trainers.py
jdrotleff: Sorry, I was wrong. It's 15 task calls, not 15 episodes. I changed it back to 4 in the docstring.

swarmrl/utils/storage_utils/trajectory_storage.py
SamTov: Thanks!

SamTov (approved):
Thanks for the quick changes! Looks good to go.

PR #115, opened 2026-05-14
feat: add continuous exploration/sampling integration and tests

jdrotleff:

* Refactored exploration policy hierarchy:
  * Added abstract `DiscreteExplorationPolicy` and `ContinuousExplorationPolicy` under `ExplorationPolicy`.
  * Updated `RandomExploration` to inherit from `DiscreteExplorationPolicy`.
  * Added and exported `GlobalOUExploration` as a continuous exploration policy using an Ornstein-Uhlenbeck process
* Refactored sampling strategy hierarchy:
  * Added abstract `DiscreteSamplingStrategy` and `ContinuousSamplingStrategy` under `SamplingStrategy`.
  * Updated `GumbelDistribution` and `CategoricalDistribution` to inherit from `DiscreteSamplingStrategy`.
  * Added `ContinuousGaussianDistribution` and exported it.
* Updated `FlaxModel` action path to support both discrete and continuous stacks:
  * Added mode inference and compatibility checks (`discrete` vs `continuous`).
  * Preserved existing discrete behavior.
  * Added continuous action branch (`sampling -> exploration`) with deployment-aware behavior.
  * Improved RNG handling with internal JAX key threading and subkey splitting.
* Added/updated tests:
  * New OU exploration tests.
  * New Gaussian sampling tests.
  * Updated Flax network tests for continuous mode + mode mismatch.
  * Updated sampling strategy base test to handle abstract base class.

Run the CI
##### Targeted tests for this PR scope
- `unit_tests/exploration_policies/test_ornstein_uhlenbeck_exploration.py`
- `unit_tests/sampling_strategies/test_gaussian.py`
- `unit_tests/sampling_strategies/test_sampling_strategy.py`
- `unit_tests/networks/test_flax_network.py`

#### Acknowledgment

This PR builds on the foundation established by @Mag-r. The architecture and implementation follow the direction set in that earlier work.
Many thanks for the contributions.

### DIFF
--- a/swarmrl/exploration_policies/ornstein_uhlenbeck_exploration.py
+++ b/swarmrl/exploration_policies/ornstein_uhlenbeck_exploration.py
@@ -1,0 +1,125 @@
+"""
+Global Ornstein-Uhlenbeck (OU) exploration for continuous actions.
+"""
+
+from typing import Any
+
+import jax
+import jax.numpy as np
+
+from swarmrl.exploration_policies.exploration_policy import ContinuousExplorationPolicy
+
+
+class GlobalOUExploration(ContinuousExplorationPolicy):
+    """
+    Adds temporally-correlated OU noise to continuous model actions,
+    as defined here https://en.wikipedia.org/wiki/Ornstein–Uhlenbeck_process#Definition:
+    dx_t = theta * (mu - x_t) * dt + sigma * dW_t. For simplicity dt=1 is assumed.
+
+    Parameters/Symbols:
+    - theta: Mean-reversion speed.
+    - mu: Long-term mean (the value the noise settles around).
+    - sigma: Volatility / scaling factor.
+    - dW_t: Wiener process increment (Standard Brownian Motion).
+        In discrete time with dt=1, dW_t is sampled from N(0, 1).
+
+    Epsilon gating is applied per action dimension.
+    For an action vector like ``[a, b, c, ...]``, each component is
+    perturbed independently with probability ``epsilon`` at every step.
+    The OU internal state is still updated every step, even when a component
+    is not applied to the output action in that step.
+
+    """
+
+    def __init__(
+        self,
+        action_limits: Any,
+        drift: float = 0.1,
+        volatility: float = 0.1,
+        long_term_mean: float = 0.0,
+        action_dimension: int = 3,
+        epsilon: float = 1.0,
+    ) -> None:
+        """
+        Initialize the OU exploration process.
+
+        Parameters
+        ----------
+        action_limits : array-like, shape (action_dimension, 2)
+            Per-action lower/upper bounds used for clipping and noise scaling.
+        drift : float
+            Mean-reversion strength (theta). Must be > 0.
+        volatility : float
+            Diffusion scale (sigma).
+        long_term_mean : float
+            Long-term mean (mu) of the OU process.
+        action_dimension : int
+            Number of continuous action dimensions.
+        epsilon : float
+            Probability of applying OU noise in a step. Must be > 0.
+        """
+        self.drift: float = float(drift)
+        self.volatility: float = float(volatility)
+        self.long_term_mean: np.ndarray = np.asarray(long_term_mean, dtype=np.float32)
+        self.action_dimension: int = int(action_dimension)
+        self.action_limits: np.ndarray = np.asarray(action_limits, dtype=np.float32)
+        self.noise: np.ndarray = np.zeros(self.action_dimension, dtype=np.float32)
+        self.epsilon: np.ndarray = np.asarray(epsilon, dtype=np.float32)
+
+        if self.drift <= 0:
+            raise ValueError("drift needs to be greater than 0")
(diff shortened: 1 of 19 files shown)


swarmrl/exploration_policies/ornstein_uhlenbeck_exploration.py
```diff
@@ -0,0 +1,125 @@
...
+
+from typing import Any
+
+import jax
+import jax.numpy as np
+
+from swarmrl.exploration_policies.exploration_policy import ContinuousExplorationPolicy
+
+
+class GlobalOUExploration(ContinuousExplorationPolicy):
```
SamTov, line 13: Could you add a paper or website reference to this sampling method? It would be nice to have somewhere to look.

swarmrl/exploration_policies/ornstein_uhlenbeck_exploration.py
```diff
@@ -0,0 +1,125 @@
...
+            * jax.random.normal(key_normal, shape=self.noise.shape, dtype=np.float32)
+        )
+
+        self.noise = self.noise + long_term_noise_shift + random_noise
+
+        # Decide whether to apply exploration in this step (per action dimension).
+        should_explore = (
+            jax.random.uniform(key_uniform, shape=self.noise.shape, dtype=np.float32)
+            < self.epsilon
+        ).astype(np.float32)
```
SamTov: Shall we have this be an argument somewhere or global parameter? We typically use float 64 for simulations. Not saying we need to use it for our ML, but I do think we should be able to control it. This may be the subject of a different PR as I think the float32 is quite widely spread.

swarmrl/networks/flax_network.py
```diff
@@ -73,17 +87,46 @@ def __init__(
...
 
             self.epoch_count = 0
 
+    @staticmethod
+    def _infer_exploration_mode(exploration_policy: ExplorationPolicy) -> str:
+        if isinstance(exploration_policy, ContinuousExplorationPolicy):
+            return "continuous"
+        if isinstance(exploration_policy, DiscreteExplorationPolicy):
+            return "discrete"
+        return "discrete"
```
SamTov: not super happy about the default discrete return. If the policy is somehow not one of the two we have implemented, I would argue we need an error raise.

swarmrl/networks/flax_network.py
```diff
@@ -178,19 +227,48 @@ def compute_action(self, observables: List):
             )
         logger.debug(f"{logits=}")  # (n_colloids, n_actions)
 
-        # Compute the action
-        indices = self.sampling_strategy(logits)
-        # Add a small value to the log_probs to avoid log(0) errors.
-        eps = 1e-8
-        log_probs = np.log(jax.nn.softmax(logits) + eps)
+        if self._sample_mode == "discrete":
```
SamTov: Okay so why can't this be completely generic and push the sampling logic to the parent classes for discrete or continuous sampling? I would argue it belongs there and not in the network class. We can check elsewhere if the sampling strategy is correct somewhere else. What do you think? Is there a way to clean this up a little bit?

SamTov (changes requested):
I have a concern about where the sampling is taking place. Please see those comments. Otherwise, looks pretty good.

swarmrl/networks/flax_network.py
jdrotleff: good catch, you're right.

swarmrl/exploration_policies/ornstein_uhlenbeck_exploration.py
jdrotleff: For this PR I would add a `float_precision` class attribute. 
I've looked into it and found that currently our value functions are using float32 explicitly. 
I think it's a good idea to add a global parameter for this in another PR, adding it to my list.

swarmrl/networks/flax_network.py
jdrotleff: You’re right, that would be cleaner architecture. I'll think about how

jdrotleff: On a second thought, if we push pure sampling to the sampling parents and pure exploration to the exploration parents, the network class still has to handle the mode-specific logic. 

For the discrete case, it would still need to manage the `softmax` and `take_along_axis` orchestration to keep the explored indices and log probs in sync, while the continuous case is just a straight pass-through. This means the network can't be truly generic this way.

If we try to avoid that by hiding the exploration inside the sampling classes to keep the network clean, we end up mixing their responsibilities.

What if we introduce an `ActionSelector` to handle this instead? This would let us clean up the network completely while keeping responsibilities strictly separated:
- `SamplingStrategy` transforms model outputs into mathematical distributions.
- `ExplorationPolicy` only explores.
- `ActionSelector` then becomes an orchestrator putting the two together. It also exposes a clean interface to the network, managing different modes internally.

What do you think about it?

SamTov: Yeah I think that's a good idea.

SamTov (approved):
Thanks of the updates! LGTM.

PR #116, opened 2026-05-19
Fix SpeciesSearch single-target different-type reward bug (Issue #93)

### LINKED ISSUE #93: Species Search not working with only one particle to be sensed
**Describe the bug**
When using the SpeciesSearch -task to train agents to find a single colloid from a different type, no training happens and the agents just act completely random. This is caused by the _compute_single_particle_task_ -function in _species_search.py_.
  
https://github.com/SwarmRL/SwarmRL/blob/eda4d9eba4273094bdba1ccd9f83ed3388a22708/swarmrl/tasks/searching/species_search.py#L126-L130

In this case _distances_ is of the shape (1,). So in line 128 the _size=distances.shape[0] - 1_ becomes 0 and _indices_ consists of just an empty array. Therefore _distances_ in line 130 also becomes an empty array on which no rewards can be calculated.

**To Reproduce**
Run SwarmRL training with a single colloid of type 1 and a bunch of agents of type 0.
Only give them the task to find the colloid of type 1.

A corresponding trainingscript is linked here:

[SpeciesSearchBug.zip](https://github.com/SwarmRL/SwarmRL/files/15456991/SpeciesSearchBug.zip)

**Expected behavior**
Normally the agents should learn to find and translate to the single colloid. Just like in the _gradient_sensing_ -task.

jdrotleff:
Fix #93 
current logic wrongly removes one element unconditionally, leaving empty distances

* Added a test for the described bug.
* Updated `species_search.py` in `compute_single_particle_task`. Removed the fragile nonzero(..., size=n-1) indexing logic and added mask-based handling

Run CI

### DIFF
--- a/.github/workflows/espresso.yml
+++ b/.github/workflows/espresso.yml
@@ -49,6 +49,10 @@ jobs:
 
       - name: Run test suite
         run: |
+          # Ensure MODULEPATH contains the ESPResSo zen2 modules before restore.
+          # This mirrors the `module use` used in the Install step so
+          # `module restore espresso` can find the saved module.
+          test "${RUNNER_ARCH}" = "X64" && module use /cvmfs/dev.eessi.io/espresso/versions/2023.06/software/linux/x86_64/amd/zen2/modules/all
           module restore espresso
           export NUM_PROC=$(nproc)
           export OMP_PROC_BIND=false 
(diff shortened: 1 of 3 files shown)


SamTov (approved):
lgtm

PR #119, opened 2026-06-05
Feature/rod torque and reward timing

BenedictRosskopf:
Fix #98, #102

**From #102 — Decouple reward calculation from action step**
* Adds `Agent.calc_reward()` and `ForceFunction.calc_reward()` as separate methods
* Reward is now computed *after* the new state is reached, fixing a timing error where rewards were evaluated on stale state

**From #98 — `RodTorque` task and torque utilities**
* Adds `RodTorque` task: rewards agents based on the torque they collectively apply to a passive rod, weighted by the rod's current angular velocity
* Adds `compute_rod_particle_distances()` and `compute_torque_on_rod()` to `colloid_utils`
* Updates `compute_forces()` to project force along the colloid's director
* Updates `compute_torque_partition_on_rod()` to accept colloid directors
* Adds `save_best_network` parameter to `EpisodicTrainer` to checkpoint the best-performing network during training
* Adds unit tests for all torque utilities and an ESPResSo integration test for cross-particle interactions


CI/espresso_tests/unit_tests/test_cross.py
```diff
@@ -0,0 +1,78 @@
...
+
+import numpy as np
+import pint
+
+from swarmrl.agents import dummy_models
+from swarmrl.engine import espresso
+from swarmrl.force_functions import ForceFunction
+
+
+class CrossTest(ut.TestCase):
```
SamTov: If you're not using the ut package for anything in the class, I would just remove it. We run the tests with `pytest` anyway.

CI/espresso_tests/unit_tests/test_cross.py
```diff
@@ -0,0 +1,78 @@
...
+            # check correct wca between differently sized particles
+            wca_params = runner.system.non_bonded_inter[0, 2].wca.get_params()
+            cutoff = wca_params["sigma"] * 2 ** (1 / 6)
+            np.testing.assert_allclose(
+                cutoff,
+                coll_rad.m_as("sim_length") + cross_thickness.m_as("sim_length") / 2,
+            )
+
+
+if __name__ == "__main__":
```
SamTov: Same here. You don't need this main code.

swarmrl/agents/actor_critic.py
```diff
@@ -213,3 +213,35 @@ def calc_action(self, colloids: typing.List[Colloid]) -> typing.List[Action]:
         self.kill_switch = self.task.kill_switch
 
         return chosen_actions
+
+    def calc_reward(
```
SamTov, line 210: Is this deliberately included in your code here? If so, the update methods needs to be included so they don't recompute the loss.

swarmrl/agents/agent.py
```diff
@@ -58,3 +58,19 @@ def calc_action(
                 Flag capable of ending simulation.
         """
         raise NotImplementedError("Implemented in Child class.")
+
+    def calc_reward(
```
SamTov, line 64: Same question.

swarmrl/force_functions/force_fn.py
```diff
@@ -84,3 +84,23 @@ def calc_action(self, colloids: typing.List[Colloid]) -> typing.List[Action]:
         self.kill_switch = any(switches)
 
         return list(actions.values())
+
+    def calc_reward(
```
SamTov, line 88: Same question

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,249 @@
...
+                Directors of the colloids.
+        colloid_positions : np.ndarray (n_rod, 3)
+                Positions of the colloids.
+
+        Returns
+        -------
+        torques : np.ndarray (n_colloids, )
+                Torques on the rod for each colloid.
+        """
+        torques = self.decomp_fn(rod_positions, colloid_directors, colloid_positions)[
```
SamTov, line 112: Can you break the newline on the parenthesis not the square bracket.

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,249 @@
...
+        # Update the historical rod director and velocity.
+        self._historic_rod_director = new_director
+        self._velocity_history_list = np.roll(self._velocity_history_list, -1)
+        self._velocity_history_list = self._velocity_history_list.at[
+            self._append_index
+        ].set(angular_velocity)
+
+        # Return the clipped average velocity.
+        if self.angular_velocity_scale > 0.0:
+            return np.clip(np.nanmean(self._velocity_history_list), 0.0, None)
```
SamTov, line 150: If you scrapped the nan from the array, you don't need nanmean anymore.

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,249 @@
...
+        else:
+            return np.clip(np.nanmean(self._velocity_history_list), None, 0.0)
+
+    def _torque_partition(
+        self,
+        colloid_torques_on_rod: np.ndarray,
+    ) -> np.ndarray:
+        """
+        Remove rewards for torque in the wrong direction.
```
SamTov, line 159: Do you want to remove rewards for torque in the wrong direction? If so, you can just clip the array, you don't need the if statement. Alternatively, you leave the negative values in and reduce the reward based on that.

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,249 @@
...
+        colloids : List[Colloid] (n_colloids, )
+                List of colloids to be used in the task.
+
+        Returns
+        -------
+        rewards : List[float] (n_colloids, )
+                Rewards for each colloid.
+        """
+        # Collect the important data.
+        rod = [colloid for colloid in colloids if colloid.type == self.rod_type]
```
SamTov:
```suggestion
        data = [(c.pos, c.director) for c in colloids if c.type == self.rod_type]
        rod_positions, rod_directors = map(np.array, zip(*data))
```

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,249 @@
...
+        rod_positions = np.array([colloid.pos for colloid in rod])
+        rod_directors = np.array([colloid.director for colloid in rod])
+
+        chosen_colloids = [
```
SamTov:
```suggestion
        data = [(c.pos, c.director) for c in colloids if c.type == self.particle_type]
        colloid_positions, colloid_directors = map(np.array, zip(*data))
```

swarmrl/utils/colloid_utils.py
```diff
@@ -41,14 +45,24 @@ def compute_forces(r):
...
+    force : jnp.ndarray (dimension, )
+        Force between the two colloids applied along the director
+        of the colloid.
     """
 
     def _sub_compute(r):
-        return 1 / jnp.linalg.norm(r) ** 12
+        return (
+            1e-8 + 1 / (jnp.linalg.norm(r) + 1e-8) ** 12
+        )  # Add epsilon numbers to avoid the gradient to be NaN
```
SamTov: Why do you need epsilon in two places? I would keep one in the denominator, but why the additional term?  The max value is capped at 1 / 1e-8 and the min value is capped at 1 / infty, which will tend to zero. The 1e-8 addition term would just be noise.

swarmrl/utils/colloid_utils.py
```diff
@@ -89,30 +103,104 @@ def compute_torque(force, direction):
...
+        Directors of the colloids.
+    rod_positions : jnp.ndarray (rod_particles, 3)
+        Positions of the rod particles.
+    rod_directions : jnp.ndarray (rod_particles, 3)
+        Directors of the rod particles.
     """
     # (n_colloids, rod_particles, 3)
     distance_matrix = compute_distance_matrix(colloid_positions, rod_positions)
-    distance_matrix = distance_matrix[:, :, :2]
+    # distance_matrix = distance_matrix[:, :, :2]
```
SamTov: Remove if not needed.

SamTov (changes requested):
Code is look good. Please check out my comments, especially those on the new `calc_loss` term. Also, I believe all tests are failing. Once that's cleaned up, we can continue.

CI/espresso_tests/unit_tests/test_cross.py
BenedictRosskopf: Removed the `unittest` import and converted the `CrossTest(ut.TestCase)` class into a standalone `test_cross()` function with plain `assert` statements, so it now runs under pytest like the rest of the suite.

CI/espresso_tests/unit_tests/test_cross.py
BenedictRosskopf: Removed the `if __name__ == "__main__"` block as well.

swarmrl/agents/actor_critic.py
BenedictRosskopf: `calc_reward()` is the new home for reward computation, called after the integrator step once the new state is available. `calc_action()` had a leftover duplicate reward-computation block from the merge that was appending to `trajectory.rewards` a second time, which would have desynced the trajectory lists used in `update_agent()`'s loss computation. I removed that duplicate block, so `calc_reward()` is now the sole place rewards get appended

swarmrl/force_functions/force_fn.py
BenedictRosskopf: Same as above

swarmrl/tasks/object_movement/rod_torque.py
BenedictRosskopf: Done

swarmrl/tasks/object_movement/rod_torque.py
BenedictRosskopf: Switched to `np.mean`

swarmrl/tasks/object_movement/rod_torque.py
BenedictRosskopf: Simplified `_torque_partition` to a single clip + sign expression, no if/else needed

swarmrl/tasks/object_movement/rod_torque.py
BenedictRosskopf: Applied both suggestions. `__call__` now collects rod and colloid data via list comprehension + `zip(*...)` unpacking

swarmrl/utils/colloid_utils.py
BenedictRosskopf: Removed the redundant `1e-8 +` term and kept only the denominator epsilon

swarmrl/utils/colloid_utils.py
BenedictRosskopf: Removed

swarmrl/trainers/episodic_trainer.py
```diff
@@ -60,6 +61,9 @@ def perform_rl_training(
                 cycle_index is passed to the EsperessoMD engine as 'h5_group_tag'. See
                 the implementation in the test_semi_episodic_data_writing function in
                 CI/espresso_tests/integration_tests/test_rl_trainers.py
+        save_best_network : str
```
jdrotleff: I think this might be an artifact of a custom checkpointer of PR #98, coming from a version without a dedicated Checkpointer (introduced in #96)
Line 129 in this file is where checkpointing is happening already, so we should not add a second, redundant one.

swarmrl/trainers/episodic_trainer.py
```diff
@@ -71,6 +75,8 @@ def perform_rl_training(
         current_reward = 0.0
         force_fn = self.initialize_training()
         cycle_index = 0
+        best_running_reward = 0.0
```
jdrotleff: Same here about the checkpointing

swarmrl/trainers/episodic_trainer.py
```diff
@@ -122,22 +128,27 @@ def perform_rl_training(
                 rewards[episode] = current_reward
                 self.maybe_save_checkpoint(rewards, episode, current_reward)
 
+                window_start = max(0, episode + 1 - 10)
```
jdrotleff: Same here about the checkpointing

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,238 @@
...
+    Push a Rod.
+    """
+
+    def __init__(
+        self,
+        rod_type: int = 1,
+        particle_type: int = 0,
+        direction: str = "CCW",
+        angular_velocity_scale: int = 100,
+        velocity_history: int = 100,
```
jdrotleff: If I read `velocity_history`, i think of an array containing the actual historical values. But this variable is used more like a "velocity history buffer size", as explained in the docstring "Number of steps to average the velocity over".

I suggest changing the variable name to something more descriptive to prevent misunderstandings.

CI/unit_tests/tasks/test_rod_torque.py
```diff
@@ -0,0 +1,158 @@
...
+        for colloid in colloids:
+            velocity = task._compute_angular_velocity(colloid.director)
+            assert velocity == pytest.approx(0.0)
+
+    def test_velocity_history(self):
+        """
+        Setup the test class.
+        """
+        task = RodTorque(
+            direction="CCW", angular_velocity_scale=1.0, velocity_history=100
```
jdrotleff: See comment in rod_torque abouot the variable name `velocity_history`. Using the same name once for an int variable and once for the actual history might be confusing.

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,238 @@
...
+        # Update the historical rod director and velocity.
+        self._historic_rod_director = new_director
+        self._velocity_history_list = np.roll(self._velocity_history_list, -1)
+        self._velocity_history_list = self._velocity_history_list.at[
+            self._append_index
+        ].set(angular_velocity)
+
+        # Return the clipped average velocity.
+        if self.angular_velocity_scale > 0.0:
+            return np.clip(np.mean(self._velocity_history_list), 0.0, None)
```
jdrotleff: The `_velocity_history_list` is initialized with only zeros. When starting a training, this will lead to a biased mean as long as the velocity history list is not completely filled with actual values from training. 

We could initialize the list as `self._velocity_history_list = np.full((self.velocity_history,), np.nan)` and use `np.nanmean(self._velocity_history_list)` instead. 

np.nanmean would then also be needed in the respective tests.

CI/unit_tests/tasks/test_rod_torque.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Test positive rewards.
+        starting_director, colloids = create_trajectory(direction_scale=1)
+        task._historic_rod_director = starting_director
+
+        # Test that the velocity is correct
+        for index, colloid in enumerate(colloids):
+            velocity = task._compute_angular_velocity(colloid.director)
+            velocity_history[index] = self.reference_velocity
+            assert velocity == pytest.approx(np.mean(velocity_history))
```
jdrotleff: See the comment at `rod_torque.py` about initializing the velocity history with `np.nan` instead of zeros. If changed, `np.nanmean` is needed here.

swarmrl/engine/espresso.py
```diff
@@ -1290,6 +1438,7 @@ def integrate(self, n_slices, force_model: ForceFunction = None):
             self.system.integrator.run(
                 steps_to_next, reuse_forces=True, recalc_forces=False
             )
+            force_model.calc_reward(self.colloids)
```
jdrotleff: `calc_reward` is executed at the bottom of the integration loop without a guard condition. As soon as `steps_per_write_interval`and `steps_per_slice` are not perfectly divisible, RL trajectories can desync.

We could add a check using the same logic as in the manage forces update, but slightly shifted.

```suggestion
            if self.step_idx == self.params.steps_per_slice * (self.slice_idx - 1):
                force_model.calc_reward(self.colloids)
```

swarmrl/engine/espresso.py
jdrotleff, line 1424: This is the context I mean in the next comment.

swarmrl/engine/espresso.py
```diff
@@ -673,6 +673,154 @@ def add_rod(
...
+        if self.n_dims != 2:
+            raise ValueError("cross can only be added in 2d")
+        if cross_center[2].magnitude != 0:
+            raise ValueError(
+                f"cross center z-component must be 0. You gave {cross_center}"
+            )
+        if n_particles % 2 != 1:
+            raise ValueError(f"n_particles must be uneven. You gave {n_particles}")
+
+        def create_orthonormal_vector(vec):
```
jdrotleff, line 751: We could move this method to the utils.

swarmrl/engine/espresso.py
```diff
@@ -673,6 +673,154 @@ def add_rod(
...
+            Creates a vector orthonormal to the given one. Notice, that the third
+            dimension should always be 0.
+            This doesn't test for zero vectors, because it is only an inner function.
+            """
+            z_vector = np.array([0, 0, 1])
+            orthonormal_vector = np.cross(vec, z_vector)
+
+            return orthonormal_vector / np.linalg.norm(orthonormal_vector)
+
+        center_pos = cross_center.m_as("sim_length")
```
jdrotleff, line 762: This code is pretty similar as the code from line 628 to 674, as a cross is almost the same as two perpendicular, merged rods.

I'm fine with keeping it as it is. I just wondered if we could somehow combine the common logic into a helper method to reduce boilerplate code.

swarmrl/force_functions/force_fn.py
```diff
@@ -84,3 +84,23 @@ def calc_action(self, colloids: typing.List[Colloid]) -> typing.List[Action]:
...
+                List of colloids in the simulation.
+        external_reward: float
+                External reward from the environment.
+
+        """
+
+        for agent in self.agents:
+            self.agents[agent].calc_reward(
+                colloids=colloids, external_reward=external_reward
+            )
```
jdrotleff, line 106: Do we need to check the kill switches here as well?
```
        switches = []
        for agent in self.agents:
            self.agents[agent].calc_reward(
                colloids=colloids, external_reward=external_reward
            )
            switches.append(self.agents[agent].kill_switch)
        self.kill_switch = any(switches)
```

swarmrl/losses/policy_gradient_loss.py
```diff
@@ -115,9 +115,11 @@ def compute_loss(self, network: Network, episode_data):
         -------
 
         """
-        feature_data = jnp.array(episode_data.features)
-        action_data = jnp.array(episode_data.actions)
-        reward_data = jnp.array(episode_data.rewards)
+        # Restructure the data to shift the rewards to after the action
+        # is taken.
+        feature_data = jnp.array(episode_data.features)[:-1]
```
jdrotleff: I don't think we need to shift the reward here, it would be correct already. In the `integrate()` method in `espresso.py` we start by choosing actions via `manage_forces()`. Then we use exactly these actions and integrate the environment. After this, we compute the rewards.
This way, the same index of the data in the `episode_data` container already corresponds to the correct tuple of s, a, r_t+1.
If we would shift it, I think we would get the tuple s_t, a_t, r_t+2.

Or did I oversee something?
If we would need to change it here, we should also take a look at the PPO loss.

swarmrl/tasks/object_movement/rod_rotation.py
```diff
@@ -49,18 +49,15 @@ def __init__(
...
         self.partition = partition
         self.rod_type = rod_type
+        self.velocity_history = velocity_history
 
         if direction == "CW":
             angular_velocity_scale *= -1  # CW is negative
 
         self.angular_velocity_scale = angular_velocity_scale
-        self._velocity_history = np.nan * np.ones(velocity_history)
+        self._velocity_history = np.zeros(velocity_history)
```
jdrotleff: Same comment as in rod_torque. I think we should keep the `np.nan` approach for correct means. Renaming `velocity_history` to something like `velocity_history_size` or similar would also be suitable here.

swarmrl/tasks/object_movement/rod_rotation.py
jdrotleff: Can you correct that?
```suggestion
        Constructor for the rod rotation task.
```

jdrotleff: typo

```suggestion
        only need to take one for the historical value.
```

swarmrl/tasks/object_movement/rod_rotation.py
```diff
@@ -108,10 +107,17 @@ def _compute_angular_velocity(self, new_director: np.ndarray):
 
         # Update the historical rod director and velocity.
         self._historic_rod_director = new_director
-        self._velocity_history = np.roll(self._velocity_history, -1)
-        self._velocity_history = self._velocity_history.at[self._append_index].set(
-            angular_velocity
-        )
+
+        # If velocity history is uninitialized (all zeros), fill with current value.
+        if np.all(self._velocity_history == 0):
```
jdrotleff: I think using `np.nan` as initial value instead of zeros would make this check redundant.
It would also be a more exact mean than using the current value as placeholder for the next `velocity_history_size` entries.

swarmrl/tasks/object_movement/rod_rotation.py
```diff
@@ -108,10 +107,17 @@ def _compute_angular_velocity(self, new_director: np.ndarray):
...
+            self._velocity_history = np.full_like(
+                self._velocity_history, angular_velocity
+            )
+        else:
+            self._velocity_history = np.roll(self._velocity_history, -1)
+            self._velocity_history = self._velocity_history.at[self._append_index].set(
```
jdrotleff: Could be `self._velocity_history.at[-1]`, right?


```suggestion
            self._velocity_history = self._velocity_history.at[-1].set(
```

swarmrl/tasks/object_movement/rod_rotation.py
```diff
@@ -79,6 +76,8 @@ def initialize(self, colloids: List[Colloid]):
         -------
         Updates the class state.
         """
+        self._velocity_history = np.zeros(self.velocity_history)
+        self._append_index = int(self.velocity_history - 1)
```
jdrotleff: This would mean it's always the last index, which allows using `velocity_history[-1]` instead of `velocity_history[self._append_index]`, wouldn't it?
```suggestion
```

swarmrl/trainers/episodic_trainer.py
```diff
@@ -122,22 +128,27 @@ def perform_rl_training(
...
+                    save_best_network is not None
+                    and running_reward > best_running_reward
+                ):
+                    best_running_reward = running_reward
+                    self.export_models(directory=f"{save_best_network}/best_network")
+
                 logger.debug(f"{episode=}")
                 logger.debug(f"{current_reward=}")
 
-                display_episode = episode + 1
```
jdrotleff, line 128: I would prefer to keep the `display_episode logic`. I've added it only recently, it was probable never part of Tobias' PR.

It shows the correct running reward right from the start, decoupled from the checkpointer logic.

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,238 @@
...
+        else:
+            return np.clip(np.mean(self._velocity_history_list), None, 0.0)
+
+    def _torque_partition(
```
jdrotleff: Based on this name, I expect a method that calculates how a total amount of torque is divided and distributed among multiple components. In contrast, torques into the wrong direction are filtered out. I suggest renaming to something like `sort_out_opposite_direction_torques`.

swarmrl/tasks/object_movement/rod_torque.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def _compute_torque_and_velocity_reward(
+        self,
+        rod_directors: np.ndarray,
+        rod_positions: np.ndarray,
+        colloid_directors: np.ndarray,
+        colloid_positions: np.ndarray,
+    ):
+        """
+        Get the torques, the turning velocity and apply the scaling.
```
jdrotleff, line 183: Because we always use the clipped torque partition, I would add this information to the docstring to be clear about it.


```suggestion
        Get the torques, the turning velocity and apply the scaling. 
        Note that only torques into the desired direction are considered for 
        reward computation.
```

swarmrl/tasks/object_movement/rod_rotation.py
```diff
@@ -49,18 +49,15 @@ def __init__(
...
         self.rod_type = rod_type
+        self.velocity_history = velocity_history
 
         if direction == "CW":
             angular_velocity_scale *= -1  # CW is negative
 
         self.angular_velocity_scale = angular_velocity_scale
-        self._velocity_history = np.nan * np.ones(velocity_history)
+        self._velocity_history = np.zeros(velocity_history)
         self._append_index = int(velocity_history - 1)
```
jdrotleff: Could be removed
```suggestion
```

swarmrl/utils/colloid_utils.py
```diff
@@ -89,30 +101,103 @@ def compute_torque(force, direction):
 
 
 @jax.jit
-def compute_torque_partition_on_rod(colloid_positions, rod_positions, rod_directions):
+def compute_torque_partition_on_rod(
```
jdrotleff: This method is called `compute_torque_partition_on_rod`, but I don't get where the partition comes from.
We also have the `compute_torque_on_rod` method, where normalization factors are used for computing the torques. This looks more partitioned to me.

Might it be that the names should be the other way around?

(Any change here will probably also interfere with the tests, we should keep this in mind.)

`compute_torque_partition_on_rod` is imported in the `rod_rotation´-task.
`compute_torque_on_rod` is imported in the `rod_torque`-task. 
If we should rename it, we should not forget renaming the imports there.

jdrotleff (changes requested):
I like your changes, thanks! 

I've come across a few points:
* I think I have found some minor artifacts from the old PRs that are outdated by the checkpointer module.
* I suggest renaming a few variables and methods for more clarity.
* The new splitted action-reward calculation might need a security check to not desyinc the RL trajectory. This might be associated with the failing tests, but I'm not sure.
* Because of the splitting, we don't need to slice the episodic_data, if I'm not wrong there.
* I suggest using np.nans for the velocity_history instead of zeros to get correct mean values. 
* Also, if we really want to fix the s,a,r tuple issue everywhere, we could have a look at the `real_experiment.py` to check if it's implemented correctly there.

Please check my comments and let me know what you think about my suggestions.

Thanks for all the work!

swarmrl/force_functions/force_fn.py
BenedictRosskopf: Tried this, but it breaks `test_continous_training_early_stop` and `test_variable_episodic_training`. Kill detection runs through `trajectory.killed`, which is only set in `calc_action (actor_critic.py)`. Setting `force_fn.kill_switch` from `calc_reward` makes the engine loop break before the next `calc_action`, so the kill never gets recorded and training runs too long. Left the kill handling in `calc_action` only.

BenedictRosskopf:
> Also, if we really want to fix the s,a,r tuple issue everywhere, we could have a look at the real_experiment.py to check if it's implemented correctly there.

`real_experiment.py` doesn't call `calc_reward` (only `calc_action`), so there should be no reward shift to fix there

swarmrl/tasks/object_movement/rod_rotation.py
BenedictRosskopf: Went with `np.nan `as initial value and the original `np.nanmean`. Reverted the change from https://github.com/SwarmRL/SwarmRL/pull/119#discussion_r3386358582; but now the nans are used

swarmrl/force_functions/force_fn.py
jdrotleff: Sounds good

jdrotleff (approved):
LGTM

Given that all open discussions with Sam will be resolved, this PR is ready to merge from my side

SamTov (approved):
Thanks for the great work! From my side, good to go!

PR #120, opened 2026-06-25
fix/ defer engine finalization to fix disk I/O training bottleneck

jdrotleff:
The `EpisodicTrainer` was explicitly calling `self.engine.finalize()` at the end of every single episode. Because `finalize()` forces an immediate disk flush of the trajectory data buffer, this completely bypassed the `write_chunk_size` setting configured in the `EspressoMD` engine.

The `ContinuousTrainer` had a structural indentation bug where `system_runner.finalize()` sat inside the active `with progress:` block context. If a run completed successfully without triggering an early stop, this might be skipped.

* Deferred the `self.engine.finalize()` call so that it triggers only when the engine object is actually about to be terminated or replaced.
* Removed the forced `finalize()` from the main episodic loop execution block.
* Added a `finalize()` catch right before `self.engine = None` during a system reset / agent killed condition, ensuring no buffered RAM data is lost.
* Added a final `finalize()` wrap-up at the very end of `perform_rl_training()` to flush the last remaining buffered chunk upon training completion.
* Unindented the fallback `finalize()` block in `ContinuousTrainer` outside of the progress bar context to guarantee execution upon normal training completion.


SamTov (approved):
Okay I had another look and it seems fine from my side. Good to go.

jdrotleff:
Nice!

PR #122, opened 2026-07-30
refactor: clarify random exploration masking

jdrotleff:
Replace the arithmetic masks with an explicit Boolean mask and `np.where`. 
The old names contradicted their behavior: `to_be_changed` actually marked actions to keep. The new logic directly expresses the intended replacement probability.


SamTov (approved):
LGTM

PR #123, opened 2026-07-30
fix: pass JAX PRNG keys directly to exploration policies

jdrotleff:
Pass JAX PRNG keys directly to exploration policies and update interfaces and tests accordingly.
Avoid converting JAX values to Python integers and keep random-key handling consistent with JAX conventions.


SamTov (approved):
Also looks good!

